PR #6, opened 2013-05-28
Swri internal mods

shaun-edwards:
Updated config/launch files to work with real robot in addition to simulated robot.


gavanderhoorn:
Seems the corresponding branch was never removed, so `swri_internal_mods` has been part of this repository (and all clones) for the past 8 years ;)

Issue #3, opened 2013-05-22
Port IKFast from Care-O-Bot and generalize to all URs


shaun-edwards:
Kelsey Hawkins (Georgia Tech) and Mathias (IPA) have created an analytic solution to the UR arm configuration.  The solution is specific to the UR10 but could easily be generalize to all URs with similar arm configurations.  

The Care-O-Bot implementation can be found here: https://github.com/ipa-mdl/cob_manipulation/blob/moveit/cob_kinematics/ikfast/src/ikfast_ur10.cpp

The KDL IK struggles to find IK solutions sometimes, so porting this code to the ROS-I version would improve functionality.

shaun-edwards:
The Care-O-Bot code is licensed LGPL.  The GT code is licensed BSD.  The kinematics code should be ported from the GT repository in order to avoid any licensing issues.  The GT code can be found here: http://code.google.com/p/gt-ros-pkg/source/browse/?repo=cpl&name=kelsey_sandbox#git%2Funiversal_robot_cpl%2Fur_kinematics%253Fstate%253Dclosed

shaun-edwards:
The ur_kinematics packages was ported from the GT code repository.  However, the code has not been wrapped in the kinematics plug in interface, so it cannot be used in arm_navigation or moveit.  The Care-O-Bot code appears to have this wrapper but is unusable due to the LGPL license.  We will have to wrap the kinematics ourselves.

shaun-edwards:
Fix merged, closing.

PR #8, opened 2013-06-04
added IKfast compatibility functions

mathias-luedtke:
in response to issue #3
proper license header is still missing, assuming BSD applies for all parts.. (@kphawkins: right?)

The code should be compatible with all available IKfast wrapper


shaun-edwards:
ur_kin.cpp comes directly from the GT google code site, which has a blanket BSD.  After I merge this, I will go ahead and add the BSD header just to be clear.

I still need a little more time to review the code in order to understand what it is doing.

PR #11, opened 2013-08-28
Update ur_driver

karlyeurl:
Added parsers for the FORCE_MODE_DATA and the ADDITIONAL_INFO (new state structures) as discussed with Renaud Detry. Fixed test_move.py as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,5 @@ universal_robot
 ===============
 
 This repository provides ROS support for the universal robots.  This repo holds source code for all versions > groovy.  For those versions <= groovy see: hg https://kforge.ros.org/ros_industrial/universal_robot
+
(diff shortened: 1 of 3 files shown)


shaun-edwards:
@gomezc A software update for the UR has broken our driver (see https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/x_XwjCiZSEo ).  This patch fixes the issue, but I want to make sure that it does not break older UR software versions (i.e. our UR).  Can you please test this.  It looks like test_move.py might be a good way of verifying the driver still works.

gomezc:
@shaun-edwards I just finished checking out the patch and everything seems to work fine with our older UR software. I ran the driver.py and the test_move.py both successfully.

abubeck:
@shaun-edwards The patch works fine on our ur10 robot since about a week together with MoveIt. So far no issues.

Issue #16, opened 2013-09-23
UR Firmware V1.8 Incompatibilities


apether:
Hi, 

I recently tried to utilise this package on a UR5 Robot running V1.8.11281 of their software which was released on Aug 07 2013. The result was very similar to that described in issue 9, caused by changes between V1.6 and V1.7 so it appears this is similar. It also reported an unknown package type, so I guess more have been added.

https://github.com/ros-industrial/universal_robot/issues/9

Has anyone else tried running the package on a V1.8 robot? This was the first time I'd tried this package, I assume I would have had more luck on a robot running V1.7. 

Here is the console output:

```
linux@ubuntu:~$ roslaunch ur_bringup ur5.launch robot_ip:=192.168.0.102
... logging to /home/linux/.ros/log/4e12e310-2033-11e3-8faa-b888e3f3f398/roslaunch-ubuntu-3991.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://ubuntu:48369/

SUMMARY
========

PARAMETERS
 * /robot_description
 * /robot_ip_address
 * /rosdistro
 * /rosversion
 * /tf2_buffer_server/buffer_size

NODES
  /
    robot_state_publisher (robot_state_publisher/state_publisher)
    tf2_buffer_server (tf2_ros/buffer_server)
    ur_driver (ur_driver/driver.py)

ROS_MASTER_URI=http://localhost:11311

core service [/rosout] found
process[robot_state_publisher-1]: started with pid [4013]
process[ur_driver-2]: started with pid [4030]
process[tf2_buffer_server-3]: started with pid [4031]
Setting prefix to 
[WARN] [WallTime: 1379489595.494830] No calibration offset for joint "shoulder_pan_joint"
[WARN] [WallTime: 1379489595.496176] No calibration offset for joint "shoulder_lift_joint"
[WARN] [WallTime: 1379489595.497629] No calibration offset for joint "elbow_joint"
[WARN] [WallTime: 1379489595.499230] No calibration offset for joint "wrist_1_joint"
[WARN] [WallTime: 1379489595.501033] No calibration offset for joint "wrist_2_joint"
[WARN] [WallTime: 1379489595.503084] No calibration offset for joint "wrist_3_joint"
[ERROR] [WallTime: 1379489595.503234] Loaded calibration offsets: {}
Disconnected.  Reconnecting
[...]

PR #23, opened 2013-10-25
Ignoring unknown pkt types, related to #16.

gavanderhoorn:
This lets the driver ignore unknown packet types, instead of bailing out.

Without rate-limiting the current implementation prints a lot of warnings, this might be nice to fix.

This is against `groovy-devel`, if that is a problem I can rebase on `hydro-devel` as well.


gavanderhoorn:
84856a2 adds some primitive rate-limiting to the printing of the warning message.

gavanderhoorn:
Ping?

Issue #29, opened 2013-11-22
ur_driver: package.xml doesn't list python module dependencies


gavanderhoorn:
From [this](http://answers.ros.org/question/103191/in-ur5_driverdriverpy-importerror-no-module-named-beautifulsoup/) ROS Answers question:

The `package.xml` file of `ur_driver` seems to be missing the Python module dependencies required for its execution.

Adding them also allows users installing from source to use `rosdep` to install all needed dependencies.

PR #30, opened 2013-11-25
Fix issue #29: missing Beautifulsoup Python dependency

gavanderhoorn:
This adds the missing `run_depend` on `python-beautifulsoup` to `package.xml` of `ur_driver`.

Also adds `run_depend` on `trajectory_msgs`, as that was missing as well.

TODO: I'm not sure the `build_depend` tags are supposed to be there. [Python module dependencies](http://docs.ros.org/groovy/api/catkin/html/howto/python_module_dependencies.html) only talks about `run_depend`. 

---

First reviewed pull-request? @sedwards?

### DIFF
--- a/ur_driver/package.xml
+++ b/ur_driver/package.xml
@@ -46,10 +46,12 @@
   <build_depend>control_msgs</build_depend>
   <build_depend>rospy</build_depend>
   <build_depend>sensor_msgs</build_depend>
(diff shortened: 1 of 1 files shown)


floweisshardt:
I fully support the idea of reviews for pull-requests. so this is my vote:

+1

Issue #38, opened 2014-02-10
ur5 moveit cfg: missing run_depend on industrial_robot_simulator


gavanderhoorn:
See ros-industrial/ros_industrial_issues#23.

PR #42, opened 2014-03-16
Fix issue #38: add missing run_depend on ind_robot_simulator

gavanderhoorn:
As per subject.

I'm not sure, but the UR5 MoveIt config package manifest might be missing some other (run|build)_depends as well.


shaun-edwards:
+1

PR #43, opened 2014-03-20
Updating the URDFs

fmessmer:
- removing some obsolete files
- update latest hydro xacro structure, i.e. xmlns
- unifying ur5 and ur10, e.g. xacro param names and arm_ee_link
- use latest transmission_interfaces


shaun-edwards:
Thanks for the contribution, just a couple of questions.

Why are the ur5_robot.urdf and ur5_joint_limited_robot.urdf obsolete?

Was the `ee_link` removed from the ur5?  This could have implications for backwards comparability.

fmessmer:
The *.urdf files have been added in association with the moveit_configs (see commit messages).
This has been done at a time where the moveit_setup_assistant did not support xacro files - which it does now.

As to the ee_link, this was mainly for consistency with the ur10. The ur10 does not have such an ee_link. 
We at IPA also require the ee_link, but as it wasn't there for the ur10, we added it into a new xacro macro together with another additional fixed base_link (there has been a previous discussion about that). 
ipa320/cob_common/blob/hydro_dev/cob_description/urdf/ur_arm/ur5_arm.urdf.xacro
ipa320/cob_common/blob/hydro_dev/cob_description/urdf/ur_arm/ur10_arm.urdf.xacro

Now, with the ur5 having an ee_link, the link is not unique anymore. Also, maybe various users will have different understandings about the "correct" orientation of such ee_link. Thus it should be added by the respective user accordingly.

What do you think?

gavanderhoorn:
Would this perhaps also be an opportunity to take ros-industrial/ros_industrial_issues#24 into account?

shaun-edwards:
@ipa-fxm, as a rule we have checked in the URDF files in addition to the xacros that generate them.  This is pretty much legacy right now, but if anybody is expecting them to be there, their code will now be broken.  Unless they are hurting anything, I don't think we should get rid of them.

As for the `ee_link`, @gavanderhoorn, provided a link to some discussion we have had on standardizing the last link (what we call `tool0`).  I'm not sure what the answer is here, but removing a link frame could cause headaches for anybody utilizing that frame.  Perhaps a good compromise is to add the `tool0` and change `ee_link` tool `deprecated_ee_link`.  This still breaks code, but it gives the user a hint as to why.

fmessmer:
I would like to comment and propose another idea later the day...

fmessmer:
Some commits have been added to this PullRequest:
- the ee_link has been added again for the ur5
- the ee_link has been added for the ur10 as well for consistency

I also added new xacro files ur5_limited and ur10_limited modelling a ur5/ur10 with reduced joint limits.

I would like to keep the old .urdf files removed as they can easily be generated by using: 
'''rosrun xacro xacro.py model.xacro > model.urdf'''

As to the tool0 discussion mentioned by @gavanderhoorn , I would consider this in a separate PullRequest and not mix it into this one. Maybe @gavanderhoorn can do the required changes and send this PR.

Any comments?

gavanderhoorn:
I'm willing to do the modifications, but I just figured this is a PR that is '_updating the URDFs_'. Adding the `tool0` frame would seem to be a part of that :), which is why I suggested it. In fact, I think the `ee_link` is the `tool0` link.

shaun-edwards:
I'm OK if `tool0` is not included in this request.  We can add a bug for that later.

I do insist that the URDF's remain in the repo.  We have several projects that reference the URDF's and not the xacro's.  By pulling down new changes or upgrading, these projects will be broken.  Worse yet, this is an issue that won't be detected until runtime.  

While I agree that they can be easily generated, I don't know that everybody would know that.  In my experience, most people do not understand and/or aren't familiar with xacros.  Unless there is a good reason to leave these out, I don't see the harm in leaving them in.

Finally, generating the xacros would have to be done within the universal package (in order for them to be located in the same package by roslaunch).  This means someone would have to generate them in the `opt/ros` directory or in a git workspace.  Either place is problematic.

fmessmer:
@shaun-edwards 
I see your point. I will re-add the URDFs again tomorrow!
Do you want me to add the exact same files which I removed, i.e. ur5_robot.urdf and ur5_joint_limited_robot.urdf?

Or should I create new URDFs from the current xacro's?
If you prefer generating new URDFs from the current xacro's, I would generate four new URDFs, i.e. ur5_robot.urdf, ur5_joint_limited_robot.urdf, ur10_robot.urdf and ur10_joint_limited_robot.urdf from the respective xacro files.
Of course, I will do the renaming from urX_limited_robot to urX_joint_limited_robot so that we will end up with the old filenames.

Just let me know which way you prefer...

shaun-edwards:
Assuming the xacros are backwards compatible (i.e. they only have added or corrected frames or data), let's run them.

fmessmer:
Generated and added the URDFs!

@shaun-edwards 
@ipa-fmw 
Happy Releasing!

floweisshardt:
visual review: OK

shaun-edwards:
Added #49 for later.

PR #44, opened 2014-03-20
Fix universal simulation in gazebo

fmessmer:
requires changes from #43
- many changes due to latest gazebo, i.e. gazebo_ros, gazebo_plugins, controller_manager


shaun-edwards:
@ipa-fmw, can you review.  I'm not familiar with the Gazebo stuff at all.

fmessmer:
please only accept this pull request after #43 is resolved

floweisshardt:
@shaun-edwards: changes regarding gazebo look fine. This will not influence anything on the real robot behaviour. Probably still needs some tuning of the gazebo controller in ur_gazebo/controller/arm_controller_ur5.yaml and ur_gazebo/controller/arm_controller_ur10.yaml.

as @ipa-fxm stated, we should wait for #43 to be solved first.

fmessmer:
Some commits have been added to this PullRequest:
I modified the launch files so that both the normal and the limited version of the ur5/ur10 (see #43 ) can be started in ur_gazebo and in ur_bringup.
This can be done using an new launch file argument "limited". By default it is set to "false" which means the normal version is started. By setting it to "true" the limited version is used. E.g.:
roslaunch ur_gazebo ur5.launch limited:=true

fmessmer:
Done!

@shaun-edwards 
@ipa-fmw 
Happy Releasing!

floweisshardt:
visual review: OK

ur_bringup/launch/ur10_bringup.launch
```diff
@@ -10,9 +10,12 @@
   
   <!-- robot_ip: IP-address of the robot's socket-messaging server -->
   <arg name="robot_ip" />
+  <arg name="limited" default="false"/>
```
gavanderhoorn, line 14: May I suggest to provide a 'convenience' launchfile, `ur10_joint_limited_bringup.launch`, which sets this argument to `true`, so users can load that directly? Ideally, the name is identical to the corresponding urdf / xacro.

In my experience users have trouble finding launchfile parameters, even if they are documented well. Providing the launchfile would make it show up immediately when they try to tab-complete.

ur_bringup/launch/ur5_bringup.launch
```diff
@@ -10,9 +10,12 @@
   
   <!-- robot_ip: IP-address of the robot's socket-messaging server -->
   <arg name="robot_ip" />
+  <arg name="limited" default="false"/>
```
gavanderhoorn, line 14: See earlier comment.

shaun-edwards:
Added issue to address @gavanderhoorn comments.  This is a usability issue and should be addressed, but not necessarily hold up this request.

shaun-edwards:
+1

PR #45, opened 2014-03-20
New moveit_configs

fmessmer:
requires #43 and #44
- re-generate X_moveit_config packages using latest moveit_setup_assistant and updated ur_description

NOTE:
Although the moveit_configs are correct, (most of the time) it is not possible to find a planned path due to the large joint value range (-2_pi...2_pi) in the ur_description.


shaun-edwards:
We have found that the moveit packages aren't very useful due to the large joint ranges as well.  The original solution was to create a moveit package using a joint limited urdf.  This worked, but wasn't ideal.  I don't think committing non-working moveit packages is a good solution either, since many people use the moveit packages to "try things out".  Is there another solution?

fmessmer:
We haven't been using moveit with the universal robots without a lot of hacks either.
Even then we had our own moveit_configs as our robots were mounted on a mobile platform.

Thus, do you want me to remove both moveit_config packages?

shaun-edwards:
I'd like to have some working moveit configs, even if it is based on the clamped joint limits.  I'm assuming the ur5 config was deleted due to the change in the URDF (i.e. the removal of `ee_link`).  I could add this one back in, assuming we are OK with the additional URDF that has the joint limits clamped at -pi to pi.

fmessmer:
I would like to comment and propose another idea later the day...

fmessmer:
I updated the moveit_configs so that they fit with the new urdf from #43 , i.e. they include the ee_link.
Similar as in #44 , a launch file argument "limited" can be used to chose between the normal and the limited robot model.
Use
'''roslaunch ur5_moveit_config demo.launch limited:=true'''
to use the demo mode with the limited version of the ur5.
Use
'''roslaunch ur5_moveit_config move_group.launch limited:=true'''
to use moveit together with simulation or real hardware for the limited version of the ur5.

fmessmer:
Done!

@shaun-edwards 
@ipa-fmw 
Happy Releasing!

floweisshardt:
visual review: OK

shaun-edwards:
+1

PR #46, opened 2014-03-20
Use global namespace for the parameters and topics in the driver

fmessmer:


shaun-edwards:
@ipa-fxm, why is this required.  Wouldn't this cause problems for multi-robot systems?

abubeck:
If you launch the ur_driver in a namespace (in our case arm_controller) the driver will expect the joint_states topic and the robot_description parameter inside of the namespace. For the joint_states topic you can write a launch file remap but you cannot do that for the robot_description.
In a multiple robot setup you probably have to use namespacing but you will still have your urdf in a central place (/robot_description).
Therefore I think it makes sense to have robot_description and joint_states as a global topic/parameter.

floweisshardt:
+1

shaun-edwards:
@abubeck, I do believe you can remap parameters names as part of the launch file (see [here](http://wiki.ros.org/Names)).  In fact for our multi-robot systems we typically have separate robot descriptions, joint state publishers, and robot state publishers, for each of our robots all under their own namespaces.  If and when we needed to access this data, launch remapping has worked.

In addition, the [driver namespace REP](http://www.ros.org/reps/rep-0135.html), while not addressing explicit global namespacing does indicate that all drivers should be launched within a namespace with all topics and parameters specified in a relative manor.

If we chose to create these topics/parameters in the global namespace, then this would not only be different from all other ROS-Industrial nodes, but most ROS nodes.  Do you know of an example of a ROS node that specifies a global namespace.

fmessmer:
We will test and discuss this and come back later the day...

fmessmer:
We tested the parameter remapping and it works for us...never used remapping for parameters before...;-)
Thus, no changes required! Closing this PullRequest!

PR #47, opened 2014-03-21
cleanup, catkin_lint and install tags

fmessmer:


floweisshardt:
looks good, I'll integrate that to the beta release

Issue #50, opened 2014-03-28
Create UR bringup launch files that automatically load joint limited URDFs


shaun-edwards:
There are two versions of URDFs for the Universal robots.  The bringup scripts automatically load the URDFs based on a `limited` argument.  If set to true the URDF with limited joint motions (-PI to PI) are loaded.  These joint limits address path planning limitations in MoveIt.

See #44 for more info.

fmessmer:
PullRequest #51 solves this issue!

fmessmer:
Please close this issue as the respective PR has been merged already!

floweisshardt:
solved by https://github.com/ros-industrial/universal_robot/pull/51

PR #51, opened 2014-03-30
convenience launch files for joint_limited versions

fmessmer:
solves ros-industrial/universal_robot/issues/50

### DIFF
--- a/ur_bringup/ur10_joint_limited.launch
+++ b/ur_bringup/ur10_joint_limited.launch
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<!--
+  Universal robot ur10 launch.  Wraps ur10_bringup.launch
+  
+  NOTE: THIS FILE IS DEPRECATED AND MAY BE REMOVED IN THE FUTURE
+
+  Usage:
+    ur10.launch robot_ip:=<value>
+-->
+<launch>
+  
+  <!-- robot_ip: IP-address of the robot's socket-messaging server -->
+  <arg name="robot_ip" />
+
+  <!-- ur common -->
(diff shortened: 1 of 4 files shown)


ur_bringup/ur10_joint_limited.launch
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<!--
+  Universal robot ur10 launch.  Wraps ur10_bringup.launch
+  
+  NOTE: THIS FILE IS DEPRECATED AND MAY BE REMOVED IN THE FUTURE
+
```
gavanderhoorn, line 6: Should we perhaps take this chance to remove the deprecated launchfiles? I often have a hard enough time figuring out which launchfile I should start now, let alone when we add an additional few.

gavanderhoorn:
+1 from me. This would fix #50.

fmessmer:
An option for solving this would be to rename launch files that should not be called directly to launchfilename.xml
However, we would than have a problem with backwards compatibility again...

gavanderhoorn:
ipa-fxm wrote:

> An option for solving this would be to rename launch files that should not be called directly to <launchfilename>.xml
> However, we would than have a problem with backwards compatibility again...

we try to adopt a tick-tock approach with ROS-I: deprecate in one
release (Groovy), remove in the following (Hydro). AFAIK the deprecation
warning has been there in Groovy, which would give us the option to
remove it in Hydro.

shaun-edwards:
Can this be merged or do we want to move deprecated launch files?

gavanderhoorn:
The current changes are ok. So this can be merged.

Perhaps the removal of deprecated files should be done in a separate PR.

But I do vote for removing the files.

PR #52, opened 2014-03-31
bring back changes from release branch

floweisshardt:
this is just bringing back changes done during the release process (adding changelogs and increasing version number)


shaun-edwards:
+1, FYI...I typically don't ask for review on these since these are really just release artifacts.

Issue #56, opened 2014-04-04
Remove deprecated launch files.


shaun-edwards:
There are several launch files that have been marked as deprecated (see #51).  We should verify that they have been marked deprecated for at least one release cycle and then remove them.

gavanderhoorn:
The current Fuerte and Groovy releases of those launchfiles don't seem to contain the deprecation warning, so it seems they can only be removed after Hydro.

Additionally, it might be a good idea to mark the launchfile as deprecated (not just as a comment in the text). Seems the launch tag [has support](http://wiki.ros.org/roslaunch/XML/launch) for that.

gavanderhoorn:
This can be closed. The files haven't been removed as the fuerte / groovy releases apparently did not warn people about the deprecation. Those warnings have been added, so Indigo can now remove the launchfiles in the root.

TODO: remove the launchfiles from the root which were added as part of the efforts by @ipa-fxm. These have never been there in earlier releases, and we don't want to add any more files to the root, so they should be removed.

Alternatively, we could retarget this issue to the Indigo release and keep this open.

fmessmer:
The newly added launchfiles for the joint_limited version have been removed from the ur_bringup root directory in PR #64

fmessmer:
The marked launch files have been removed in PR #90

gavanderhoorn:
Seeing as we now have an `indigo-devel` branch, could this issue be retargeted for Indigo?

fmessmer:
This issue can be closed as the launch files were removed in PR #90 
See https://github.com/ipa-fxm/universal_robot/commit/ffee2b831fc4c159bc82b26d3ca1387631539bb1

gavanderhoorn:
Yep, hadn't seen that commit. Tnx.

PR #57, opened 2014-04-07
For issue #56: add deprecation warning to deprecated launch files

gavanderhoorn:
As discussed in #56, this results in a warning message being printed to the console whenever these launch files are used.

I propose to also delete the joint limited launchfiles from the root of the package: I copied them to the `launch` subdirectory where I think they should be (ie: new files, not deprecated and part of the default set of files). I can add a commit removing those in the root if desired.


floweisshardt:
visual review: OK

gavanderhoorn:
Do we want to remove the `*_joint_limited` files from the root as well?

Issue #60, opened 2014-04-16
description: urdfs are out-of-date wrt xacros


gavanderhoorn:
See also #58: apparently somewhere during the merging of the latest set of pull requests, the updates to the urdfs were not properly merged. They are now not in-synch with the current xacros.

This should be fixed, as people expect those to describe the same robot.

fmessmer:
This issue can be closed after PR #63 is merged

PR #63, opened 2014-04-17
updated urdf files

fmessmer:
solves ros-industrial/universal_robot/issues/60

### DIFF
--- a/ur_description/urdf/ur10_joint_limited_robot.urdf
+++ b/ur_description/urdf/ur10_joint_limited_robot.urdf
@@ -6,22 +6,24 @@
 <robot name="ur10" xmlns:xacro="http://www.ros.org/wiki/xacro">
   <!-- common stuff -->
   <gazebo>
-    <plugin filename="libgazebo_ros_controller_manager.so" name="gazebo_ros_controller_manager">
-      <alwaysOn>true</alwaysOn>
-      <updateRate>1000.0</updateRate>
+    <plugin filename="libgazebo_ros_control.so" name="ros_control">
+      <!--robotNamespace>/</robotNamespace-->
+      <!--robotSimType>gazebo_ros_control/DefaultRobotHWSim</robotSimType-->
     </plugin>
-    <plugin filename="libgazebo_ros_power_monitor.so" name="gazebo_ros_power_monitor_controller">
+    <!--
+    <plugin name="gazebo_ros_power_monitor_controller" filename="libgazebo_ros_power_monitor.so">
       <alwaysOn>true</alwaysOn>
       <updateRate>1.0</updateRate>
       <timeout>5</timeout>
       <powerStateTopic>power_state</powerStateTopic>
       <powerStateRate>10.0</powerStateRate>
-      <fullChargeCapacity>87.78</fullChargeCapacity>
+      <fullChargeCapacity>87.78</fullChargeCapacity>     
       <dischargeRate>-474</dischargeRate>
       <chargeRate>525</chargeRate>
       <dischargeVoltage>15.52</dischargeVoltage>
       <chargeVoltage>16.41</chargeVoltage>
     </plugin>
+-->
   </gazebo>
   <!-- ur10 -->
   <material name="UR/Blue">
--- a/ur_description/urdf/ur10_robot.urdf
+++ b/ur_description/urdf/ur10_robot.urdf
@@ -6,22 +6,24 @@
 <robot name="ur10" xmlns:xacro="http://www.ros.org/wiki/xacro">
   <!-- common stuff -->
   <gazebo>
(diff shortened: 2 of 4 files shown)


fmessmer:
Close issue #60 after this PR is merged

mikepurvis:
I suggest that these URDFs be created from the xacros during build, rather than being in the source tree, see: http://wiki.ros.org/xacro#Building_from_CMakeLists.txt

A practical example of this is in my FLIR PTU driver: https://github.com/ros-drivers/flir_ptu/blob/master/flir_ptu_description/CMakeLists.txt

Note that there's a small bug with this which prevents it working in Hydro at the moment, but it will be resolved with the next release, which I'm hoping will be soon.

shaun-edwards:
I have pushed back against creating URDFs as part of the build process since xacro is not part of the build system/tools.  Worse yet, xacro is dependent on roslaunch (for argument substitution), which is definitely not a build tool.  However, manually creating them results in the issue we currently have.  I see them as equally bad.

@ipa-fxm, I leave it up to you to decide.

+1 from me either way.

mikepurvis:
> ... since xacro is not part of the build system/tools.

I'm not sure I understand. If you have a package with a `<build_depend>` on xacro, that still won't bring it onto user machines as a runtime dependency, so users can't possibly be impacted.

The point of catkin's CFG_EXTRAS scheme is making all packages to be able to contribute build-time macros for exactly this type of scenario.

shaun-edwards:
> > ... since xacro is not part of the build system/tools.
> 
> I'm not sure I understand. If you have a package with a on xacro, that still won't bring it onto user machines as a runtime dependency, so users can't possibly be impacted.

My comment wasn't about the user being impacted.  It's more a general philosophy that build tools (i.e. anything that produces a build artifact) should not be tied to ROS releases.  For example, catkin and bloom have tried/are trying to be independent.  The compiler is certainly independent.  This independence is better for stability and limits the introduction of bugs.  

I'm particularly sensitive about URDFs since they form the basis for moveit configs and fastik solutions.  Even if URDFs are dynamically updated, these other packages are not.  By generating URDFs as part of the build process, this opens us up to bugs introduced by changes in `xacro` and `roslaunch` that create inconsistent (i.e. different) URDFs.  Without checking URDFs into the repo, we cannot catch these changes.  There aren't any build tests for this either.  Historically this is why we have committed the actual URDFs.

shaun-edwards:
@ipa-fxm , ping

fmessmer:
Well, I haven't tested the generation that @mikepurvis suggested.
As he is speaking of a bug (what's the problem?), I think that the generation is _currently_ not an option, right?

I might have said this before, I can say that URDFs are no longer needed for the moveit configs. I don't know how the FastIK has been generated and whether URDFs are still needed there (@kphawkins might have a word here). Besides, the ur_kinematics have not been tested after the recent changes within ur_description. Theoretically it should still work...

So maybe in the long run (indigo?), we actually should get rid of these (how can we mark URDFs as deprecated?). For now, we could use the one produced by this PR. In case the generation works, we could think about using this solution. @mikepurvis could you send a PR for this?

gavanderhoorn:
@ipa-fxm: I think you're missing the point @shaun-edwards is trying to make: the urdfs weren't included because we didn't know how to generate them from xacros as part of the build, or because certain pkgs needed them (moveit cfgs fi), but because generating them automatically at buildtime skips a step in our (admittedly rather manual) Q&A process.

As there are no tests available that can catch changes in urdf generation, potentially we could be releasing a support pkg that contains a faulty urdf (or worse: a working one, but with subtle differences). Because of it being automated, this would also mean that we wouldn't know about this until that pkg has been synced and distributed to all our users. The invisibility of this becomes even worse when we don't use those urdfs ourselves anymore, as is now the case with the newer moveit cfgs.

The idea with ROS-Industrial (at least AFAIU) is to provide a certain level of stability, once a package has been released. Relying on automated generation of such a crucial part of your system -- without any checks on that process in place -- would seem a bit strange.

floweisshardt:
following this controverse discussion I am suggesting the following solution:
- we primarily use the xacros (as URDFs are not used in any launch file or other package)
- for backward compatibility we'll keep the URDFs in the package
- the URDFs should be manually generated as soon as the xacros are updated (this will be the responsibility of the maintainer)
- as soon as the xacros are updated the maintainer needs to make sure that the moveit and ikfast configurations will be updated too

This way all Q&A is in the hand of the maintainer and we're not depending of automatic generation/dependencies to other tools.

shaun-edwards:
@ipa-fmw, +1, Can this be merged then?

fmessmer:
No objections against merging...

gavanderhoorn:
@ipa-fmw; essentially keep the established workflow, but try to use the xacros as much as possible. +1 from me then.

Issue #39, opened 2014-02-10
Multi-arm support


jeff-o:
Just a quick question.  I'm not able to easily test it out myself, but does the package support multiple arms?  Naturally I'd create a custom launcher that separately names the topics published by the two arms.

shaun-edwards:
@jeff-o, launching the nodes in different namespaces would only work(simply) if the two arms had separate controllers.  I'm not sure what the dual arm setup looks like.  I would recommend sending an email to the dev list (more people watch that): swri-ros-pkg-dev@googlegroups.com

jeff-o:
Ah, there's me not being specific enough.  Yes, of course there is one arm per controller in my planned configuration.  I couldn't think of any reason why it shouldn't work, but thought I'd ask anyway.

gavanderhoorn:
Is the `ur_driver` node capable of dealing with multiple connections to the same serversocket? Or should multiple instances of that node be started? In the latter case, I think the port  `ur_driver` opens should be configurable via a launch file parameter (hard coded to be `50001` now)?

I've seen at least one [ros answers](http://answers.ros.org/question/129911/dual-arm-ur10-connection-with-universal-robot-stack/) question about this, and that person started hardcoding a different `REVERSE_PORT`.

mikepurvis:
(I'm a colleague of @jeff-o's at Clearpath)

Yes, we ran into the REVERSE_PORT issue as well. Our PC had two NICs, so our solution was simply to move one arm to a different subnet on the other NIC and then change the TCPServer code to only bind to the one interface rather than all interfaces. An obvious way to avoid the hack would be to bind TCPServer to port 0, then read out the assigned port and inject it into the prog script, ~~which could perhaps be templated using empy.~~ **I now see that prog is already being templated for the host/ip of the PC running the driver, so additionally passing in the port should be very simple.**

Thanks for the awesome job done on ur_description, by the way. Our robot description turned out to be really simple, this was basically it:

```
<?xml version="1.0"?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="husky">
  <xacro:include filename="$(find husky_description)/urdf/base.urdf.xacro" />
  <xacro:include filename="$(find ur_description)/urdf/ur5.urdf.xacro" />
  <ur5_robot prefix="left_arm_" />
  <ur5_robot prefix="right_arm_" />

  <joint name="base_to_left_arm" type="fixed">
    ...
  </joint>
[...]

PR #65, opened 2014-04-23
Issue: classnames in driver

fmessmer:
this solves #2

PR #66, opened 2014-04-23
Issue: multi-arm support

fmessmer:
Solves #39

### DIFF
--- a/ur_bringup/launch/ur_common.launch
+++ b/ur_bringup/launch/ur_common.launch
@@ -11,16 +11,18 @@
 -->
 <launch>
   <!-- robot_ip: IP-address of the robot's socket-messaging server -->
-  <arg name="robot_ip" />
+  <arg name="robot_ip"/>
+  <arg name="reverse_port" default="50001"/>
   
   <!-- copy the specified IP address to be consistant with ROS-Industrial spec.
   NOTE: The ip address is actually passed to the driver on the command line -->
   <param name="/robot_ip_address" type="str" value="$(arg robot_ip)"/>
+  <param name="/robot_reverse_port" type="int" value="$(arg reverse_port)"/>
   
   <node name="robot_state_publisher" pkg="robot_state_publisher" type="state_publisher" />
 
   <!-- driver -->
-  <node name="ur_driver" pkg="ur_driver" type="driver.py" args="$(arg robot_ip)" output="screen" />
+  <node name="ur_driver" pkg="ur_driver" type="driver.py" args="$(arg robot_ip) $(arg reverse_port)" output="screen" />
    
   <!-- TF Buffer Server -->
   <node pkg="tf2_ros" type="buffer_server" name="tf2_buffer_server">
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -28,7 +28,7 @@ prevent_programming = False
 joint_offsets = {}
 
 PORT=30002
-REVERSE_PORT = 50001
+DEFAULT_REVERSE_PORT = 50001
 
 MSG_OUT = 1
 MSG_QUIT = 2
@@ -670,10 +670,12 @@ def main():
         parser.error("You must specify the robot hostname")
     elif len(args) == 1:
         robot_hostname = args[0]
-        reverse_port = REVERSE_PORT
+        reverse_port = DEFAULT_REVERSE_PORT
     elif len(args) == 2:
         robot_hostname = args[0]
         reverse_port = int(args[1])
+        if not (0 <= reverse_port <= 65535):
+                parser.error("You entered an invalid port number")
     else:
         parser.error("Wrong number of parameters")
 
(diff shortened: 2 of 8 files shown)


ur_driver/src/ur_driver/driver.py
```diff
@@ -664,11 +664,18 @@ def main():
...
-    if len(args) != 1:
+    if len(args) < 1:
         parser.error("You must specify the robot hostname")
-    robot_hostname = args[0]
+    elif len(args) == 1:
+        robot_hostname = args[0]
+        reverse_port = REVERSE_PORT
+    elif len(args) == 2:
+        robot_hostname = args[0]
+        reverse_port = int(args[1])
```
gavanderhoorn, line 676: Should we perhaps add some range checking here? To make it fail early in case the user specifies an invalid port?

ur_driver/src/ur_driver/driver.py
```diff
@@ -664,11 +664,18 @@ def main():
...
-    parser = optparse.OptionParser(usage="usage: %prog robot_hostname")
+    parser = optparse.OptionParser(usage="usage: %prog robot_hostname [reverse_port]")
     (options, args) = parser.parse_args(rospy.myargv()[1:])
-    if len(args) != 1:
+    if len(args) < 1:
         parser.error("You must specify the robot hostname")
-    robot_hostname = args[0]
+    elif len(args) == 1:
+        robot_hostname = args[0]
+        reverse_port = REVERSE_PORT
```
gavanderhoorn: Rename to `DEFAULT_REVERSE_PORT`?

fmessmer:
Basic testing with a single UR10: able to move arm by sending follow_joint_trajectory and receive joint_states...

fmessmer:
Updated this PR due to changes in PR #65

shaun-edwards:
@ipa-fxm, please resond/address to @gavanderhoorn's comments.

ur_driver/src/ur_driver/driver.py
fmessmer: I think the name really does not matter, thus I kept the current variable name

ur_driver/src/ur_driver/driver.py
fmessmer, line 676: @gavanderhoorn What would be a proper range for the ports?
Besides, I would see this as an enhancement once someone is acutally using this in a multiarm setup.

gavanderhoorn, line 676: > What would be a proper range for the ports?

Unsigned 16 bit integer, in principle.

> Besides, I would see this as an enhancement once someone is acutally using this in a multiarm setup.

I strongly disagree: I'd say checking your input for acceptable values is simply part of a proper implementation of the desired functionality.

ur_driver/src/ur_driver/driver.py
gavanderhoorn: Well, given that the changes you propose introduce a configurable port number, I think renaming this variable does matter, as this one is used as a constant (and not to store the user specified port).

It is also _only_ used in case the user does not specify a different port number to use, which implies that this is the _default_, so imho it would make sense to name it as such.

gavanderhoorn:
ping?

fmessmer:
I am currently working on other (non-ur) issues.
Also, I actually don't have a dual arm setup for further testing and improving this.
I just did the changes to allow the concepts being discussed in the related issue #39 
Feel free to send any improvments as another PR superseeding this one.

fmessmer:
@abubeck provided a solution for the suggestions given by @gavanderhoorn 
+1 from my side!

abubeck:
@ipa-fxm : README needs to be updated with port option

fmessmer:
- reverse_port can now be set through launch file arguments
- README is updated

ur_bringup/launch/ur_common.launch
```diff
@@ -11,16 +11,18 @@
 -->
 <launch>
   <!-- robot_ip: IP-address of the robot's socket-messaging server -->
-  <arg name="robot_ip" />
+  <arg name="robot_ip"/>
+  <arg name="reverse_port" default="50001"/>
   
   <!-- copy the specified IP address to be consistant with ROS-Industrial spec.
   NOTE: The ip address is actually passed to the driver on the command line -->
   <param name="/robot_ip_address" type="str" value="$(arg robot_ip)"/>
```
floweisshardt, line 19: @abubeck shouldn't this parameters be in the local namespace of each node instance?

gavanderhoorn, line 19: afaik the ROS-I standard approach is to use namespaces for that. In any other case I'd agree though.

fmessmer, line 19: It seems that this parameters isn't used at all!
(At least not for setting the robot_ip and reverse_port in the driver)
But, as I am not absolutely certain about this, I left it - for compatibility reasons....

abubeck, line 19: @ipa-fmw : thanks for the feedback. As @gavanderhoorn wrote, the way the parameters are loaded is ROS-I standard compatible. Therefore I will merge this PR as is.

PR #69, opened 2014-05-08
Issue/effort limits

fmessmer:
The effort limits in the URDFs seem to be too small to properly set up controllers in simulation.
Given the large inertias currently used in the URDFs the small effort limits are too small to be able to move the arm and receive a successful result from the follow_joint_trajectory action.
As I don't know the proper inertias, increasing the effort limits was the easiest fix for now.

At least it can be used in simulation with this PR.
We still can lateron adjust inertias, effort limits and other dynamic parameters in a next iteration.

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -77,7 +77,7 @@
     <child link = "${prefix}shoulder_link" />
     <origin xyz="0.0 0.0 ${shoulder_height}" rpy="0.0 0.0 0.0" />
     <axis xyz="0.0 0.0 1.0" />
-    <limit lower="${-2.0 * pi}" upper="${2.0 * pi}" effort="330.0" velocity="2.16"/>
+    <limit lower="${-2.0 * pi}" upper="${2.0 * pi}" effort="1000.0" velocity="2.16"/>
     <dynamics damping="1.2" friction="0.0"/>
   </joint>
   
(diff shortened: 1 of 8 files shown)


shaun-edwards:
I'm not sure where the inertia parameters came from, but I believe @kphawkins updated parameters based on the controller.  Are we sure we want to change values for which we have some basis.

kphawkins:
That controller has no feedback linearization, as far as I'm aware, so any
attempt at accurate dynamic modeling will be extremely poor. For Gazebo,
you should just have it tracking whatever gets commanded as accurately as
possible.

I'd recommend adding an alternate model which just removes the effort
limits or sets them really high. This could be accomplished with little
duplication using xacro macros. I agree with Shaun that the default should
be based on reality, or at least what we believe it to be.

shaun-edwards:
@ipa-fxm, ping

abubeck:
If I see it right we would then have 3 different urdf setups for each UR arm:
- original settings
- reduced joint limits for MoveIt
- higher effort limits for gazebo controllers
  Am I right?

Instead of creating yet another urdf file I would prefer to have one urdf and work with xacro properties and set the properties within the launch file. See:
http://answers.ros.org/question/38956/pass-parameters-to-xacro-in-launch-file/

If you agree I would go forward with this next week and update the PR.

fmessmer:
This PR is out-dated. 
A better solution is provided in #84 

Closing!

Issue #54, opened 2014-04-01
Hand teaching and pendant manual motion disabled by ROS driver.


shaun-edwards:
The ability to perform manual moves via the teach pendant or move the robot by dragging the end-effector is disabled by the ROS driver.  This capability is convenient and should not be disabled when under ROS control.  Other ROS-I drivers allow control via the teach pendant.  This is actually implemented at the controller level, but functionally the UR should act the same.

jeff-o:
On a related note, using the arrows to move the arm from the Pendant "Move" tab results in the arm moving only a few cm before stopping.  The arm can be repeatedly "jogged" to a position, but it's annoying at the very least.

ghost:
Has there been any progress on this issue?

I'm a beginner with ROS and ur_driver, but I've found that messing with the timeout values in ur_driver.py sometimes makes it work, but I'm not sure how to proceed.

fmessmer:
As @paturdc mentioned on the ros-industrial mailinglist (https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/5KBACyQNJro), he probably found a way to solve this.

@paturdc could you create a PullRequest with your changes, so that your solution can be discussed here and tested by other users? It will also give this issue some new stimulus...

de-vri-es:
Changing the timeout before the script is re-uploaded will not fix the issue. At best it will make it occur less frequently.

The problem is that any action on the pendant interrupts the current UR script. The python side detects this and re-uploads the script and by that interrupts the action you started from the pendant. Increasing the timeout might give you enough time to finish whatever movement you were trying to do from the pendant, but only if you're lucky.

What would help is if the python side didn't re-upload the script until it is actually needed, to execute a trajectory. Then the feedback on port 30002 should be used most of the time, so it might be worth it to always use that feedback and remove the feedback from the UR script (except for a heartbeat to detect the connection dropping).

ghost:
Changing the timeout is just a hack I came up with, because I really needed
both *ur_driver *and manual control at the same time. If it is not a
workable solution, I'll just keep it in my private files, for when it is
needed.

   What would help is if the python side didn't re-upload the script until it
   is actually needed, to execute a trajectory. Then the feedback on port
   30002 should be used most of the time, so it might be worth it to always
   use that feedback and remove the feedback from the UR script (except for a
   heartbeat to detect the connection dropping).

Would that solution mean the driver stops transmitting _/joint_state_ ?

[...]

PR #71, opened 2014-05-28
Fixes for the python driver to work wtih new firmware again

mathias-luedtke:
- Fix should be backward-compatible, but was not tested on old firmware yet
- Python package name is needed for working with the release version, otherwise  importing deserialize fails

### DIFF
--- a/ur_driver/src/ur_driver/deserialize.py
+++ b/ur_driver/src/ur_driver/deserialize.py
@@ -10,6 +10,7 @@ class PackageType(object):
     CONFIGURATION_DATA = 6
     FORCE_MODE_DATA = 7
     ADDITIONAL_INFO = 8
+    CALIBRATION_DATA = 9
 
 class RobotMode(object):
     RUNNING = 0
@@ -260,6 +261,8 @@ class RobotState(object):
                 rs.force_mode_data = ForceModeData.unpack(package_buf)
             elif ptype == PackageType.ADDITIONAL_INFO:
                 rs.additional_info = AdditionalInfo.unpack(package_buf)
+            elif ptype == PackageType.CALIBRATION_DATA:
+                pass # internal data, should be skipped
             else:
                 rs.unknown_ptypes.append(ptype)
         return rs
--- a/ur_driver/src/ur_driver/deserialize.py
+++ b/ur_driver/src/ur_driver/deserialize.py
@@ -187,13 +187,31 @@ class ForceModeData(object):
          fmd.robot_dexterity) = struct.unpack_from("!IBddddddd", buf)
         return fmd
 
-class AdditionalInfo(object):
+class AdditionalInfoOld(object):
     __slots__ = ['ctrl_bits', 'teach_button']
     @staticmethod
     def unpack(buf):
-        ai = AdditionalInfo()
-        (_, _, ai.ctrl_bits, ai.teach_button) = struct.unpack_from("!IBIB", buf)
+        ai = AdditionalInfoOld()
+        (_,_,ai.ctrl_bits, ai.teach_button) = struct.unpack_from("!IBIB", buf)
         return ai
+        
+class AdditionalInfoNew(object):
+    __slots__ = ['teach_button_enabled','teach_button_pressed']
+    @staticmethod
+    def unpack(buf):
+        ai = AdditionalInfoNew()
+        (_,_,ai.teach_button_enabled, ai.teach_button_pressed) = struct.unpack_from("!IBBB", buf)
+        return ai
+        
+class AdditionalInfo(object):
+    @staticmethod
+    def unpack(buf):
+        ai = AdditionalInfo()
+        (plen, ptype) = struct.unpack_from("!IB", buf)
+        if plen == 10:
+            return AdditionalInfoOld.unpack(buf)
+        else:
+            return AdditionalInfoNew.unpack(buf)
 
 class RobotState(object):
     __slots__ = ['robot_mode_data', 'joint_data', 'tool_data',
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -16,7 +16,7 @@ from sensor_msgs.msg import JointState
 from control_msgs.msg import FollowJointTrajectoryAction
 from trajectory_msgs.msg import JointTrajectory, JointTrajectoryPoint
 
-from deserialize import RobotState, RobotMode
+from ur_driver.deserialize import RobotState, RobotMode
 
 prevent_programming = False
 


shaun-edwards:
It feels like we keep chasing down these issues every time universal adds/changes a message structure.  I thought we addressed this with #23.  As far as I understand the driver should throw out a warning but continue to work.

Also, what is the _new_ firmware version?

gavanderhoorn:
@shaun-edwards: the changes in #23 allow the driver to ignore _new_ unknown packets, they don't deal with changes to already implemented ones. The changes by @ipa-mdl suggest that the `AdditionalInfo` packet has been changed.

mathias-luedtke:
I got these issues with version 1.8.11997 on a new UR5, but I am not sure when the changes were introduced.
The history in the support wiki suggests a change in September/October 2013.
#23 stopped the driver from crashing on unknown messages, but even with the message throttling the logs got filled up, so I added the new package type to compensate. However, the warnings are really helpful to discover that something has changed..
The structure of AdditionalInfo was changed fundamentally, so I had to branch based on the package length to support both versions.

shaun-edwards:
The "additional packet" was added in PR #11.  The [discussion](https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/x_XwjCiZSEo) attached to the PR indicates the change was required due to changes between Universal version 1.6 and 1.7.  It now appears that the _additional info_ packet has changed between 1.7 and 1.8.

I'd like to capture this information directly in the code (in a comment) so that future developers will understand where this ugliness came from.

We should also test backwards compatibility in 1.7 and <1.7.  I can cover <1.7, does anybody have a universal robot with 1.7 who can test this fix?  Also, we should verify that the 1.7 in fact has the _old_ additional data packet.

gavanderhoorn:
Do we actually need to be able to deserialise all packets in the stream? It might be possible to just ignore everything _but_ joint state messages.

Isn't the data being sent on port 30002 only used in case the driver isn't running on the UR controller? So that we can report joint states even without our urscript running?

mathias-luedtke:
@gavanderhoorn: good point! At the moment the driver does not expose any other data, so we could just skip the rest based on the package lengths. In order to support upcoming features of the driver (e.g. IO handling) we should establish some sort of deserialization plug-in mechanism with the joint state package handler as an example implementation.

gavanderhoorn:
@ipa-mdl: yes, skipping unknown / unneeded packets by using the embedded lengths was what I was thinking. Even so though: I think it would be good to look at whether we actually need to be able to report state when our driver is _not_ running on the controller, cause that would make the whole deserialisation of the UR protocol unnecessary.

de-vri-es:
Why use the packet length? Packets also have a type (1 for Joint Data). That should be much more reliable (and sane).

mathias-luedtke:
@gavanderhoorn: the robot mode package data is currently used to determine if the robot can be programmed. We should keep the deserialization of relevant data. For now it would work without, but  future version might run the script only for trajectory execution (I was thinking of #54).

@de-vri-es: No, the (sub) package length (sent by the UR controller) is the only reliable run-time data we get since the structure of the sub-packages (like joint data etc.) changes from time to time without further notification

de-vri-es:
@ip-mdl: If that is true then there is no reliable run-time data since any other packet type might chance to be the same length as the join data. However, as it is, every single (sub) packet starts with its length and its type (4 bytes and then 1). Are there any versions of the firmware where this is not true?

mathias-luedtke:
@de-vri-es: There is no point in guessing the package length based on the sub-package type if we can just read it from the controller as part of the same package header. But let's keep the focus on the main topic..

de-vri-es:
@ipa-mdl: I wasn't suggesting guessing the length based on the type, I was merely suggesting identyfing joint data sub-packets based on their type rather than their length. Of course you still have to use the packet length, but using only the length as a means of guessing the packets type is just as strange (and fragile) as using the type to guess the length.

mathias-luedtke:
@de-vri-es: I guess we misunderstood each other.. we were talking about skipping unused data (determined by the package type) simply by the package length instead of deserializing it completely (same as for unknown data).

de-vri-es:
@ipa-mdl: Ah, I was under the impression that you wanted to use the packet length as basis for the decision to skip data or not. That would not be a good approach since it makes it impossible to log the fact that joint data packets have changed size (if they ever alter those). But if you just inted to use the length for determining how much data to skip, then I agree completely.

gavanderhoorn:
Right, everyone ok again? :)

@ipa-mdl wrote:

> the robot mode package data is currently used to determine if the robot can be programmed. We should keep the deserialization of relevant data. 

Sure, completely dropping everything is not a good idea. Skipping non-relevant parts could allow us to work around the ever-changing payload issue.

> For now it would work without, but future version might run the script only for trajectory execution (I was thinking of #54).

I'm not sure whether we'd want to fix #54 by stopping / starting the driver all the time. I can imagine trying to better handle the various states the controller can be in (ie: manual jogging, backdriveable, etc) and implementing special logic for those states.

de-vri-es:
@gavanderhoorn: There's not really that much choice. The moment you perform any action to control the robot from the teach pendant, the currently running script is aborted. If you re-upload the driver script to start it again, whatever action you started with the teach pendant is interrupted. So in order to work together nicely with the teach pendant, the only option is to not re-upload the driver script every time it stops.

de-vri-es:
Unless of course you can detect when the user is interacting with the arm itself, but then you'd have to parse a lot more of the feedback on port 30001 again to figure out when to stop uploading the script.

mathias-luedtke:
As @de-vri-es already pointed out, the actions on the teach panel stop the driver script, so the driver has to reupload it anyway, but that discussion belongs to #54
I just wanted to point out, that we still need the data from the socket..

mathias-luedtke:
Since this is a pull request:
1. Short-term: Please test my version with older firmware to confirm that this bugfix doesn't break anything else
2. Mid-term: We should open a new issue to discuss maintenance overhead due to the protocol changes and possible solutions to this
3. Long-term: Better integration with the teach panel commands

shaun-edwards:
@jrgnicho, can you test this change against the SwRI system.  Please verify our software version on the UR as well (I believe it's 1.6).

jrgnicho:
I just ran the test_move.py script as described in https://github.com/ipa-mdl/universal_robot/tree/hydro-devel on a ur controller with the 1.6 software version and it behaved normally, meaning that it ran through a series of predefined joint values in a loop.  Unless further testing is advised the changes in this pull request can be merged safely.
+1

shaun-edwards:
It looks like we have tested this on version 1.6 and 1.8.  Does anybody have version 1.7 to test.  If not, then I recommend we move forward based on code review.

fmessmer:
Yes, I can confirm that this works on a UR10 with software version 1.6.9496
Unfortunately, we don't have 1.7.X around...

abubeck:
Ok then I agree with @shaun-edwards regarding the 1.7 testing and merge this PR.

PR #73, opened 2014-06-06
update README

fmessmer:


shaun-edwards:
In other ROS-I packages we created `moveit_planning_execution.launnch` files as described in this section of the MoveIt package generation [tutorial](http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot#Update_Configuration_Files).  This allows for a single launch to handle simulated and real robot control.  I thought this was in there already.  This launch file is lost sometimes when the moveit packages are regenerated.

+0 on this change.  I'd like to see the single launch file that is similar to other ROS-I packages, but I don't want to stand in the way of better documentation.

fmessmer:
I added the [ur5|ur10]_moveit_planning_execution.launch files, which should fix the discrepancy between the namespaces for the follow_joint_trajectory action for the REAL and SIMULATED robots.

I also used the MoveItSimpleControllerManager instead of the Pr2MoveItControllerManager

Can someone test this on the REAL robot?

As a note:
The new launch file hides quite a lot of arguments that where offered by move_group.launch, however, I think, it doesn't make much sense to have them all in the [ur5|ur10]_moveit_planning_execution.launch files again as they become more and more complex...

ur10_moveit_config/launch/ur10_moveit_planning_execution.launch
```diff
@@ -0,0 +1,25 @@
+<launch>
+  <arg name="sim" default="false" />
+  <arg name="limited" default="false"/>
+  <arg name="debug" default="false" />
+  
+  <!-- Launch moveit -->
+  <group if="$(arg sim)">
+    <!-- Remap follow_joint_trajectory -->
+    <remap from="/follow_joint_trajectory/cancel" to="/arm_controller/follow_joint_trajectory/cancel"/>
```
shaun-edwards: I believe remapping can happen at the namespace level (i.e. `remap from="/follow_joint_trajecto" to="/arm_controller/follow_joint_trajectory"`, or perhaps even simpler).  Not a show stopper, just a suggestion.

ur10_moveit_config/launch/ur10_moveit_planning_execution.launch
```diff
@@ -0,0 +1,25 @@
...
+    <remap from="/follow_joint_trajectory/feedback" to="/arm_controller/follow_joint_trajectory/feedback"/>
+    <remap from="/follow_joint_trajectory/goal" to="/arm_controller/follow_joint_trajectory/goal"/>
+    <remap from="/follow_joint_trajectory/result" to="/arm_controller/follow_joint_trajectory/result"/>
+    <remap from="/follow_joint_trajectory/status" to="/arm_controller/follow_joint_trajectory/status"/>
+    <include file="$(find ur10_moveit_config)/launch/move_group.launch">
```
shaun-edwards: It looks like this include happens regardless of the `sim` argument.  It doesn't appear that either launch will start the universal driver (maybe I am wrong).

fmessmer:
Did some simplifications...(but it was working before as well ;-)

abubeck:
Since @ipa-fxm tested this PR on the robot and in simulation + integrated shaun's suggestions I will merge this.

PR #75, opened 2014-06-17
added timeout "hacks" to allow manual motion

ghost:
We have two robots, on running on a little older software, the other with brand new. In the older one (a UR5), simply changing line 718 to 

```
r = getConnectedRobot(wait=True, timeout=50.0)
```

worked really well, although the timeout value can be adjusted up or down. In the newer robot (a UR10), the above solution does not work. However, since I HAD to make it work quickly, I found that simply adding 

```
print "Timeout for 10 s"
time.sleep(10)
```

below 

```
except EOF, ex:
```

worked well enough for practical purposes, although it's not very pretty.

Changing "time.sleep(10)" to a more elegant solution that waits for the robot to become available again, would go a long way toward solving this issue, if i understand it correctly. As I am writing my Master Thesis ATM, I don't really have time to come up with a better method.


tnl-woi:
Thanks for contributing to the repository. However, this pull request contains many changes that should **NOT** be merged into the ros-industrial repository, IMHO. For example, swap files are added and default values (e.g. the IP address of the robot) are changed, which affects other users.
About the timeout hacks, I don't understand exactly which problem they should solve. And as your description already mentions a better solution, I think it would be better to wait for someone to implement that solution than to merge this. If this pull request would really solve a generic problem, only the relevant changes should be merged.

fmessmer:
This PR was never meant to be merged....it mainly was for having a chance to look at the code of @paturdc 
It is related to (still unsolved) issue #54

tnl-woi:
@ipa-fxm Thanks for the clarification, and the reference to the issue. It makes more sense now :-)
Should this merge request be closed then? Regards, Wouter.

abubeck:
#163 works well for the use case. 
For further discussion can be found in #54

PR #77, opened 2014-06-25
Issue/tcp force

Equanox:
For some applications it could be useful to have the force or torque impacting on the end effector of the robot arm. I added a publisher with a geometry_msgs/Wrench to the driver. The published force and torque are in the base coordinate system of the robot.

Hint: The computed force is dependant on the mount of the robot. There's a possibility to set the mounted pose during runtime. I'll commit this functionality in a later pull request.

### DIFF
--- a/ur_driver/prog
+++ b/ur_driver/prog
@@ -33,6 +33,7 @@ def driverProg():
       q = get_joint_positions()
       qdot = get_joint_speeds()
       tau = get_joint_torques()
+      tcp_force = get_tcp_force()
       enter_critical
       socket_send_int(MSG_JOINT_STATES)
       socket_send_int(floor(MULT_jointstate * q[0]))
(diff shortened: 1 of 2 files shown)


fmessmer:
looks good to me
(although I haven't tested it with hardware and I currently don't have an application that would be using this feature)

gavanderhoorn:
imo it would be nicer to add a new message type to the driver's protocol (ie something like: `MSG_TCP_FORCE`) and add an explicit deserialisation case to the `while` in `handle()` in the Python driver. Instead of piggy-backing this data onto the `MSG_JOINT_STATES` msg type.

Rationale: better seperation of concerns, less ambiguity, higher testability, better backwards compatibility.

So +1 for making more sensor data available, -1 for the current implementation.

Equanox:
@gavanderhoorn you're right. I implemented the suggestions + added the WrenchStamped msg.  I think naming it MSG_WRENCH is more clean due to the ros geometry_msg/Wrench. Although i didn't choose a sequential number for the messages due to my PR #78 .

gavanderhoorn:
Does ROS have any best practices for naming topics related to forces / torques?

Other than that this now looks much better. I was wondering why you didn't use the same recipe for implementing this as you did with the payload msg type you added :).

fmessmer:
The message defines got the same value as the I/O interface mentioned in #79 
Might lead to problems in case the I/O interface makes it into this repo....

abubeck:
Looks good. I'm not aware of any best practice regarding naming. Having a topic named wrench seems fine. When we are discussing a possible extension of the ROS-I driver specification this can also be a point of standardisation.
I will merge this PR as is.

gavanderhoorn:
Shouldn't this have been tested on at least 1.6, 1.7 and 1.8?

abubeck:
This was tested on our UR's. But we don't have all versions for testing. If it doesn't work for you please add an issue.

Equanox:
It was tested  with Polyscope 1.6 and 1.7 .

gavanderhoorn:
Ok, great. Thanks. Does this mean that 1.8 hasn't been tested?

Equanox:
I had no opportunity to test it on 1.8. But while comparing the documentations (<=1.7 and 1.8)  i figured out that the torque on <=1.7 is given in N/rad and in 1.8 in Nm (the documentations says 'around these axes measyred in Newton times Meters'). Maybe this can be mentioned in the description.

From my point of view the implementation should work on 1.8 (of course without testing).

PR #78, opened 2014-07-01
Issue/heavy payload

Equanox:
When lifting heavy payloads with the ur5/ur10 the controller tends to trigger a "forced stop". To avoid  this it is necessary to set the correct payload so that the controller can predict the appearing forces due to the payload.

In this PR a new class called URServiceProvider is implemented as  interface for sending parameters to the robot during runtime.

The payload can be set between [0.0; 20.0] . It is necessary to set the correct payload when working with the force/torque from https://github.com/ros-industrial/universal_robot/pull/77


fmessmer:
The message defines got the same value as the I/O interface mentioned in #79 
Might lead to problems in case the I/O interface makes it into this repo....

abubeck:
I agree with @gavanderhoorn . Also please rebase from hydro-devel.

Equanox:
Some comments of @gavanderhoorn got lost. I think this is due to the rebase. I'll add them here to keep them in mind.

Typo: bounce -> bounds
Also: might be nice to print the bounds as part of the error.
Also2: perhaps make bounds configurable? rosparam?

fmessmer:
The message defines have been changed for the IO interface...number 8 is reserved for the set_payload message from this PR...

As the IO interface (#81 and #92) have the message definitiions in a new package `ur_msgs`, maybe we want to move the SetPayloard.srv definition to that package as well...

I suggest to deal with the recommended changes for this PR once the IO interface PRs made it into `hydro-devel`...

fmessmer:
#102 rebases this PR

abubeck:
Closing because of #102

PR #81, opened 2014-07-08
added IO capabilities

airplanesrule:
The IO modifications should be in the IO_testing branch.

### DIFF
--- a/ur_driver/prog
+++ b/ur_driver/prog
@@ -9,6 +9,11 @@ def driverProg():
   MSG_STOPJ = 6
   MSG_SERVOJ = 7
   MSG_WRENCH = 9
+  MSG_SET_DIGITAL_OUT = 10
+  MSG_GET_IO = 11
+  MSG_SET_FLAG = 12
+  MSG_SET_TOOL_VOLTAGE = 13
+  MSG_SET_ANALOG_OUT = 14
   MULT_wrench = 10000.0
   MULT_jointstate = 10000.0
   MULT_time = 1000000.0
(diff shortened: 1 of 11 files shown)


ur_driver/prog
```diff
@@ -7,6 +8,11 @@ def driverProg():
   MSG_WAYPOINT_FINISHED = 5
   MSG_STOPJ = 6
   MSG_SERVOJ = 7
+  MSG_SET_DIGITAL_OUT = 8
+  MSG_GET_IO = 9
+  MSG_SET_FLAG = 10
+  MSG_SET_TOOL_VOLTAGE = 11
+  MSG_SET_ANALOG_OUT = 12
```
fmessmer: we need to find an agreement with message gets which number
see https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur_driver/prog#L11
and https://github.com/ros-industrial/universal_robot/pull/78/files#diff-b2f4c5403df32a364117b89312e34ab7R11

fmessmer:
Please merge with the latest hydro-devel.

Some more notes:
- I somehow like the concept of having a ur_msg package as I have seen it in @wcaarls original version. What about you?
- Many messages have the same types but only different names for the message fields...can we combine and generalize them?

airplanesrule:
I updated the message numbers so I don't interfere with other people's work, added the ur_msgs folder, and made all messages either a Analog or Digital type. I think I managed to merge my changes with the hydro-devel branch, but I am new to git so I may well have messed up. Let me know if anything else should be changed.

dpsolomon:
ping?

jrgnicho:
I just tested this branch and found out that I had to make some changes in order to get it to work.  The changes address the following:
- builds ur_msgs messages and service
- modify the imports in the driver.py to use the definitions in ur_msgs

I created a branch with those changes here https://github.com/jrgnicho/universal_robot/tree/IO-Fixed

fmessmer:
In case PR #92 gets accepted this PR is obsolete and can be closed...(all changes from this PR are included in #92 )!

abubeck:
Now continued with #92

PR #82, opened 2014-07-11
UR10 URDF Update + Fixed Link Lengths

### LINKED ISSUE #74: Verify LinkLengths
I opened this issue in order to record the discussion on the mailinglist on github as well:
https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/1e3A_1BTQWQ

kphawkins:
This updates the UR10 model description to one which is based on Collada meshes.  Furthermore, the kinematic description is modified to rely solely on Denavit-Hartenberg parameters found in urcontrol.conf.  Currently, the kinematic paramaters in the UR10 urdf disagree with those in the ur_kinematics/ur_kin.h files.

Hopefully this will fix #74 and #4.


kphawkins:
I'd also like to comment that a dummy link has been added so that `base_link` can have no mass.  This should make MoveIt more happy.

shaun-edwards:
ping.

kphawkins:
Can I get someone else to test and confirm or should I just merge?

fmessmer:
I will have a look at it...however, I probably won't be able to do it before next week...as I also would like to verify that we finally get a model in this repo that works fine in simulation as well...i.e. also account for the recent issue on the mailing-list regarding the controller parameters...
I'll post my comments here!

gavanderhoorn:
@ipa-fxm: could we perhaps review this PR considering just the updated meshes and urdf? The simulation issue is preferably solved separately from this.

@kphawkins: I like the changes in this PR. One thing I noticed when testing this out was that the MoveIt config seems to be missing some info. MoveIt complains about `Could not identify parent group for end-effector 'moveit_ee'`.

fmessmer:
Some more remarks:
- Tabs vs. Spaces!
- Shall the stl meshes be removed as they are not needed anymore?
- The cube.dae file is not used. thus, it should be deleted
- You are using the scale attribute within the mesh tag, I'd prefer to correctly scale the meshes directly

Also, see my comment on https://github.com/gt-ros-pkg/universal_robot/commit/4119b726fc1adeb268f5c02628efc6f2c5ca2f1b

dpsolomon:
The kinematics look correct now, good job @kphawkins. 
- It looks like the ur10_joint_limited_limited_robot.urdf and ur10_robot.urdf still need to be updated from their corresponding xacro files.
- Also, I agree with @ipa-fxm that the mesh scale should be correct.
- If any work is done to resize the meshes, please make sure that all face normals are organized coherently; some of the triangles have flipped normals which cause display issues (I think collision still works fine).
- Eventually, better collision models should be generated as per http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Collision_STLs. Some of the links could be convex hull, but that may be inappropriate for the 2 arms as it may remove usable space.

**I think the importance of this PR trumps changing the solid models at this time. These and other minor formatting issues should be addressed, but the major changes should make their way to the ros-industrial repo ASAP. Thanks Kelsey!**

fmessmer:
When I spawn the new model in Gazebo (roslaunch ur_gazebo ur10.launch paused:=true, i.e. not actually starting the controllers), it looks like this:

screenshot_ur10_update

I have seen the flying cover of the shoulder before, but it looks ok in rviz and I don't know what's wrong. Also, I don't know why wrist_2 got another colour...as the texture looks ok e.g. in Blender...Anyone knows what the problem is?

A question: 
Is there a convention about where the origin of a mesh should be?
In this PR the origins are all wrt to a common frame (base_link?) and a origin tag is needed in the visual and collision tags...

You can also see the centers-of-gravity in the screenshot (purple boxes)...they don't seem to be correct...

Sorry for having so many issues about this PR...but I think we should do it right this time!

gavanderhoorn:
@ipa-fxm wrote:

> Is there a convention about where the origin of a mesh should be?
> In this PR the origins are all wrt to a common frame (base_link?) and a origin tag is needed in the visual and collision tags...

I think we don't have any documents that specifically state this, but it seems to be best practice to always have the origin of meshes coincide with the origin of the frame they are attached to. Or at least, that's what I've always understood. That allows all origins to be `0 0 0` triplets.

> Also, I don't know why wrist_2 got another colour...as the texture looks ok e.g. in Blender...Anyone knows what the problem is?

I can't tell too well from the screenshot, but inverted normals can affect the appearance of meshes, including colour (influences lighting). @dpsolomon mentioned that some meshes have problems with the normals.

dpsolomon:
One more important issue! The ur10.urdf.xacro is updated, but the ur10_joint_limited.urdf.xacro has not been updated. The files should be the same except that the latter has all +/- 2pi joint limits set to +/- pi so that the planners don't choke up.

This should get corrected, and the corresponding urdfs should get updated (see my previous post) before accepting PR.

kphawkins:
> @kphawkins: I like the changes in this PR. One thing I noticed when testing this out was that the MoveIt config seems to be missing some info. MoveIt complains about Could not identify parent group for end-effector 'moveit_ee'.

I wasn't the one who originally added this.  I don't know the pros/cons really.

> However, I think the link should not be added here but within the ur10_robot.urdf.xacro.

Also, it might be considerable to call this link "world", as this implicitly would attach the robot to the gazebo world by a fixed joint, thus making it less wobbly...
Done.

> Tabs vs. Spaces!

Removed all tabs from UR10 files.

> Shall the stl meshes be removed as they are not needed anymore?

Removed.

> The cube.dae file is not used. thus, it should be deleted

Removed.

> You are using the scale attribute within the mesh tag, I'd prefer to correctly scale the meshes directly
> 
> I think we don't have any documents that specifically state this, but it seems to be best practice to always have the origin of meshes coincide with the origin of the frame they are attached to. Or at least, that's what I've always understood. That allows all origins to be 0 0 0 triplets.

I don't intend to update the meshes since I don't really see a huge benefit.  It took me a while to make them and it will take more time to update them further.  If someone wants to take it upon themselves, it seems that the 3 requests are to 1) center the meshes at the origin, 2) scale them to meters instead of millimeters, and 3) clean up the normals.  This should definitely be addressed in a new issue/PR.

> It looks like the ur10_joint_limited_limited_robot.urdf and ur10_robot.urdf still need to be updated from their corresponding xacro files.

Done.

In addition, I have removed the incredibly redundant ur10_joint_limited.urdf.xacro file and instead added a xacro parameter to the ur10.urdf.xacro file.

I think I'm done addressing the issues here. Can someone pull the latest changes, confirm it doesn't break anything big, and let me know if anything else needs to be updated?

abubeck:
This was tested on UR10 with version v1.6 and v1.7. I will merge this pull request.

PR #83, opened 2014-08-01
MoveIt! plugin to replace KDL kinematics with analytic version

kphawkins:
This adds a Kinematics adapter which uses the ur_kin.cpp IK as opposed to the KDL kinematics.  The code is largely copied from the KDL implementation, but reimplements the `searchPositionIK` method and adds to the `init` method to use the ur_kin solutions.  See the `searchPositionIK` method to see how a solution is determined.

This implementation should support active joints before the base_link and after the wrist_3_link.  This is supported by first trying the supplied joint estimate and checking for an IK solution for the arm.  If there is a solution for the arm, I'll return that with the joints unchanged.  If no IK solution is found for the arm, the other joints are set randomly and the IK repeats.  The first found IK solution is returned.  In future versions, I expect to adjust how the solution is determined to optimize over a methodically sampled set of solutions.

This code includes the changed from the branch `ur10_urdf_update`.

IMPORTANT: There is a recently submitted patch in KDL necessary for this code to work properly.  Please update Orocos KDL before running this, or use the KDL_OLD_BUG_FIX flag during compilation.


abubeck:
@kphawkins : can you please rebase. Is the KDL issue solved?

fmessmer:
I did the rebase in PR#96
Also, I updated the urX_moveit_configs according to the latest models...

abubeck:
Will be continued in #96

PR #84, opened 2014-08-04
UR10 gazebo controller update

fmessmer:
This PR updates the dynamic model of the UR10 and adjusts the PID controller gains for the JointTrajectoryController used with Gazebo

builds on top of PR https://github.com/ros-industrial/universal_robot/pull/82


gavanderhoorn:
More an insight-question for me I guess: are the mass, inertia and cog values from the `urcontrol` dir on the controller not usable here (ie: is that the reason you chose to approximate those values)?

Same goes for the `max_velocity` and `max_acceleration` values. The various `.conf` files on the controller seem to contain all the relevant information?

fmessmer:
Well, the cog values that where used by Kelsey had the centers of gravity at some strange positions (see my screenshot here: https://cloud.githubusercontent.com/assets/508006/3769489/98e88686-18e1-11e4-9972-48a0cac327a3.png).
Also, the values that I extracted from the urcontrol of our UR10s did not place the centers of gravity at better positions (they where outside the meshes, too).
Thus, I think there is some transformation wrong...

And then I thought, approximating the inertias by equally distributed cylinders is still better than having '1.0' on the inertia diagonal...but I kept the ur_control values for mass and lengths.

I guess it's undoable to perfectly model a robot's dynamics for Gazebo (as friction is officially not supported)...but the behavior with these values is quite good already...you can test it together with MoveIt! and you will see that both the error while moving as well as at the goal position is less than 0.05 rad per joint...

(Besides the fact that when starting Gazebo, the controllers are not yet active and then the arm starts to drop...that's something you cannot avoid as Gazebo physics is always up first...)

fmessmer:
As to max_velocity, I used the same values for MoveIt! as they are used in the URDF..which are the same as in urcontrol...
The max_acceleration values from urcontrol (to me) seemed to be too high...but I guess the effect of the acceleration limits on the behaviour is not too big anyway...

gavanderhoorn:
@ipa-fxm wrote:

> Well, the cog values that where used by Kelsey had the centers of gravity at some strange positions (see my screenshot [here](https://cloud.githubusercontent.com/assets/508006/3769489/98e88686-18e1-11e4-9972-48a0cac327a3.png)).
> Also, the values that I extracted from the urcontrol of our UR10s did not place the centers of gravity at better positions (they where outside the meshes, too).
> Thus, I think there is some transformation wrong...

@abubeck wrote in #74 he contacted UR with some questions. Would it perhaps be an idea to ask them for some more information on how the parameters in `urcontrol` should be understood? If it is true that all robots have different parameters, understanding the `urcontrol` file would make it possible / easier to write a generator tool for customised urdfs.

> And then I thought, approximating the inertias by equally distributed cylinders is still better than having '1.0' on the inertia diagonal...but I kept the ur_control values for mass and lengths.
> 
> I guess it's undoable to perfectly model a robot's dynamics for Gazebo (as friction is officially not supported)...but the behavior with these values is quite good already...you can test it together with MoveIt! and you will see that both the error while moving as well as at the goal position is less than 0.05 rad per joint...

Please don't understand my comments as critiques in any way. I was simply trying to understand why these values did not match those in the urcontrol files (and being a bit jealous that the UR has those values 'in the open' :)).

> As to max_velocity, I used the same values for MoveIt! as they are used in the URDF..which are the same as in urcontrol...

True, my mistake.

> The max_acceleration values from urcontrol (to me) seemed to be too high...but I guess the effect of the acceleration limits on the behaviour is not too big anyway...

`joint_size1_CB3_rev1.conf`:

```
[..]
a_max = 40.0
[..]
```

I see what you mean. That doesn't really seem to make sense.

Sticky281:
I just tried running this PR with simulation and MoveIt and I'm not seeing any planners or interactive markers in RVIZ. In the ur10_urdf_update branch it still worked. Any idea why?

fmessmer:
I have added another "endeffector" group to get rid of the Warning when MoveIt! is started.
Maybe this new group is now selected as default within rviz.
Try to select "manipulator" within the PlanningRequest section of the Display settings for the MoveIt! RViz plugin...

Sticky281:
Ah yea that was it, thanks! Simulation seems to work well too.

gavanderhoorn:
@ipa-fxm wrote:

> I guess it's undoable to perfectly model a robot's dynamics for Gazebo (as friction is officially not supported).

I don't quite understand this. The Gazebo wiki seems to state the opposite: [wiki/Tutorials/1.9/friction](http://gazebosim.org/wiki/Tutorials/1.9/friction).

arennuit:
@gavanderhoorn: good point, I am also interested in knowing some more about what is meant by

```
 **I guess it's undoable to perfectly model a robot's dynamics for Gazebo**
```

This actually relates to http://answers.ros.org/question/188953/real-vs-simu-params-and-gains-behavior-difference/

abubeck:
As this pull request is an improvement to the current state wrt. kinematics and simulation and was tested with UR10 and UR5 on different controller versions I will merge this pull request.
Dynamic modelling needs to be investigated further and should be handled in a separate pull request. Any help is appreciated.

arennuit:
I follow up on the conversation about _the robot's dynamic modeling_ in [issue #87](https://github.com/ros-industrial/universal_robot/issues/87).

PR #85, opened 2014-08-05
Fixed wobbling of the UR5 (UR10 untouched).

arennuit:
The default parameters for the UR5 (dynamic params and PID gains) cause the robot to be wobbly, shaky and not very stable. I updated the parameters so the simulation behaves nicely. There is certainly much better to be done (the tracking seems to be laggy), but this is already a much better start. Note that the robot needs to be fixed to the world - which is not the case by default.

The rough updated parameter values were initially provided by jalfonso in http://answers.ros.org/question/157261/manipulator-shakeswobbles/ and refined by me using dynamic reconfigure.

Thanks,

Antoine.

### DIFF
--- a/ur_description/urdf/ur5.urdf.xacro
+++ b/ur_description/urdf/ur5.urdf.xacro
@@ -66,7 +66,7 @@ center_of_mass = [ [0, -0.02561, 0.00193], [0.2125, 0, 0.11336], [0.11993, 0.0, 
     </collision>
     <inertial>
       <mass value="${base_mass}" />
-      <inertia ixx="1.0" ixy="0.0" ixz="0.0" iyy="1.0" iyz="0.0" izz="1.0"/>
+      <inertia ixx="0.002" ixy="0.0" ixz="0.0" iyy="0.02" iyz="0.0" izz="0.002"/>
     </inertial>
   </link>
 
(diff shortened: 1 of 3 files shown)


fmessmer:
Thanks for the PR!
However, the models are currently updated for both UR5 and UR10.
Can you test #86 and let us know whether this version solves your wobbly Gazebo behaviour?

arennuit:
Sorry this is a dumb question but I am not really used to this: shall I just clone fork ipa-fxm/universal_robot.git and branch to ur5_urdf_update or is their a way to patch PR86 onto my local clone of ros-industrial/universal_robot.git (branch hydro-devel)?

gavanderhoorn:
@arennuit: one additional way would be:

```
# in your local clone
git checkout -b pr86 hydro-devel
curl https://github.com/ros-industrial/universal_robot/pull/85.patch | git am
```

Your `pr86` branch should now contain the commits in this PR. Alternatively you could indeed just add the fork of @ipa-fxm as a remote of your local clone and checkout the `ur5_urdf_update` branch.

fmessmer:
I guess the least confusing way would be to cleanly check out my branch. In your local copy of universal_robot do the following:

```
git remote add origin-ipa-fxm https://github.com/ipa-fxm/universal_robot.git
git fetch -p --all
git checkout -b pr86 origin-ipa-fxm/ur5_urdf_update
```

arennuit:
Thanks Gijs and Felix, I have tested the new params for UR5 and they are a real improvement compared to the _hydro-devel_ branch. My params were yet less wobbly but I need to compare their impact on the tracking speed - will let you know if something comes out of my tests. Thanks.

arennuit:
This PR is no longer relevant as the matter is addressed in PR86.

PR #86, opened 2014-08-05
Beautification, Unification and UR5 updates

fmessmer:
Mainly this PR aims for a unification and beautification of the UR10 and UR5 models.
(tabs vs. spaces, same structure, cleanup)

It includes #82 and #84.

The kinematical values for the UR5 have not been changed as they were correct already.

With accepting this PR, the kinematic models for both UR5 and UR10 should be conform with the actual hardware, i.e. urcontrol, also the gazebo simulation is actually usable do to updated gazebo controller parameters.

Have fun!


arennuit:
In file ur5.urdf.xacro, the following properties are no longer used, should they be removed?

 property name="shoulder_cog" value="0.0 0.00193 -0.02561"
 property name="upper_arm_cog" value="0.0 -0.024201 0.2125"
 property name="forearm_cog" value="0.0 0.0265 0.11993"
 property name="wrist_1_cog" value="0.0 0.110949 0.01634"
 property name="wrist_2_cog" value="0.0 0.0018 0.11099"
 property name="wrist_3_cog" value="0.0 0.001159 0.0"

fmessmer:
No, they should not be removed...those values are from the actual universal_robot control and should actually be used for the inertia tags...
However, it is currently unclear what the reference frame is for these transformations...
Thus, until this is clearified the properties just stay in the xacro as an unused placeholder/reminder

arennuit:
I have another proposition for the gains values in _arm_controller_ur5.yaml_ (yet less wobbling, much reduced overshoot and no observed impact on tracking speed). Should I push this commit in repo_https://github.com/ipa-fxm/universal_robot.git_ | branch _ur5_urdf_update_ so you can include it in the pull request?

The updated values are:

```
shoulder_pan_joint: {p: 10000.0, i: 500.0, d: 200.0, i_clamp: 100.0}
shoulder_lift_joint: {p: 10000.0, i: 500.0, d: 200.0, i_clamp: 100.0}
elbow_joint: {p: 10000.0, i: 500.0, d: 100.0, i_clamp: 100.0}
wrist_1_joint: {p: 100.0, i: 0.0, d: 10.0, i_clamp: 0.0}
wrist_2_joint: {p: 100.0, i: 0.0, d: 10.0, i_clamp: 0.0}
wrist_3_joint: {p: 100.0, i: 0.0, d: 0.0, i_clamp: 0.0}
```

By the way what is the difference between _i_clamp_ and _i_clamp_min/max_ which was previously used? From the source code http://docs.ros.org/indigo/api/control_toolbox/html/pid_8cpp_source.html i_clamp_min/max do not seem to be used...

gavanderhoorn:
The manifest of `ur_gazebo` is missing some run/build_depends:
- `pr2_mechanism_diagnostics`
- `gazebo_ros_pkgs`
- `gazebo_ros_control`
- `ros_controllers`

These should be added, as I ran a `rosdep check` and was under the impression I had all the prerequisites installed, but the launch still failed.

I'm not sure whether these are the proper top-level dependencies to be added to the manifest, but I could at least get things to run with these.

fmessmer:
@gavanderhoorn, I added the run_dependencies
@arennuit, your PIDs seem to be fine as well

fmessmer:
@arennuit 
By setting `i_clamp`, `i_clamp_min` and `i_camp_max` are set automatically using `i_clamp` with accoring sign

abubeck:
As this was tested and all comments are resolved, I will merge this PR.

PR #88, opened 2014-08-08
Enhancement/update meshes

fmessmer:
This PR updates the UR5 and UR10 meshes according to ROS-Industrial standards as described here:
http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot

This included:
- setting the origins of the meshes into the according joint frame
- proper scaling and actual application of position, rotation and scale within the meshes
- removal of additional light and cameras in mesh files
- use convex hulls for collision tags
- remove gazebo materials form URDF description
- unify color and texture of ur5 and ur10

The meshes have been re-worked by Feyyaz Cakir of Fraunhofer IPA. Thank you!

I also re-generated the plain URDF files to reflect the latest updates in the XACRO files

NOTE: The kinematic and dynamic description is untouched by this PR


shaun-edwards:
I'm having some trouble verifying this change.  When I try to run either the ur10 or ur5 MoveIt config package, I get the following:

```
ros@ROS:~/ros/universal/src/universal_robot/ur_description/urdf$ roslaunch ur5_moveit_config ur5_moveit_planning_execution.launch 

...


process[move_group-2]: started with pid [4798]
[ERROR] [1407858395.155613913]: Robot model not loaded
[ERROR] [1407858395.183974780]: Planning scene not configured
[move_group-2] process has finished cleanly
log file: /home/ros/.ros/log/d661ab00-2237-11e4-b5fb-080027f3243a/move_group-2*.log
```

This also occurs on the hydro-devel branch, so it's not caused by this change.  Is preventing me from verifying the fix.  Is anybody else seeing this problem?

gavanderhoorn:
I see the same behaviour, but I think this is not necessarily wrong, just different. I think you are expecting `urN_moveit_planning_execution.launch` to behave like the ones described in the ROS-I tutorial. Afaict, the ones in the UR MoveIt config are supposed to be included into other launch files that take care of the planning scene and the `robot_model`.

gavanderhoorn:
@ipa-fxm: could you comment on the quality of the visual meshes (faces, vertices)? Loading `demo.launch` takes quite some time on my machine, perhaps some of the meshes should be decimated?

fmessmer:
Hmm, I don't have any problems using the urN_planning_execution.launch 
Did you run it as is described in the README https://github.com/ros-industrial/universal_robot/blob/hydro-devel/README.md?

Maybe we should review the urN_moveit_config package structure after the ur_description has been verified.
They need an update anyway due to new collision meshes, Also we could test @kphawkins analytic IK solver plugin. Then we could also make it conform to the tutorial http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot
For now you should be able to use the urN_moveit_config package by starting the move_group.launch and the moveit_rviz.launch files

As to the meshes:
I am quite sure we did not modify the complexity of the meshes. We only changed the origin and the textures... So they should not have more vertices as the ones that were used before our PR.
Here are some numbers from the largest meshes respectively:
- ur5/visual: UpperArm.dae has 14728 vertices and 29436 faces
- ur5/collision: UpperArm.dae has 2870 vertices and 5736 faces
- ur10/visual: Wrist2.dae has 18694 vertices and 37373 faces
- ur10/collision: Wrist2.dae has 2873 vertices and 5542 faces

I recognize that the UR10 meshes are more detailed as they e.g. have the screws fixing the caps modeled as well as the UR symbol on those caps.

Still the complexity and size of the meshes seems to be ok to me.

shaun-edwards:
I figured out what I was doing wrong (not reading the whole README).  I'm about to release a package that automates much of the process for creating standard ROS-I packages for different robots.  I would wait until that is released before we tackle the UR.

I've loaded up demo.launch for both the UR5 and 10.  In a VM, with software rendering, it doesn't feel any less responsive then other moveit packages I have used.

fmessmer:
so, what does this mean for this PR?
"Go merge" or "Hold"?

shaun-edwards:
I would decimate the visual models if it is easy (I don't have much experience with dae files).  @gavanderhoorn, you make the call.

gavanderhoorn:
@shaun-edwards wrote:

> I've loaded up demo.launch for both the UR5 and 10. In a VM, with software rendering, it doesn't feel any less responsive then other moveit packages I have used.

The delay is in the loading: if you have a decent PC the rendering overhead should not be noticeable.

Lets keep the meshes as they are: I'm more worried about the 'floating cap' that is somewhere in the air if you start the gazebo sim.

fmessmer:
The "floating cap" is fixed with the meshes in this PR.
@gavanderhoorn, are you using the correct meshes?

UR5:
ur5
UZR10
ur10

gavanderhoorn:
@ipa-fxm: yes I am, I was just trying to post something relying on my memory (very bad idea). Afaicr this seems ok then. I haven't noticed anything strange about the meshes (although I can only check so much here).

gavanderhoorn:
Just checked again and I noticed that some meshes have holes in them. See the screenshots. This is only in the UR10 I think (I checked the UR5 and couldn't find any):

holes1

holes2

fmessmer:
@gavanderhoorn the meshes don't have holes. those face seem to have a wrong face normals (pointing inwards rather than outwards)
There is a feature in Blender called "recalculate" (normals) but it does not seem to help repair the face normals....I also tried meshlab...but I am not able to flip the faces coherently...
However, I think it's fine like this....so my suggestion is to merge as is!

gavanderhoorn:
@ipa-fxm wrote:

> the meshes don't have holes. those face seem to have a wrong face normals (pointing inwards rather than outwards)

Aren't faces with inverted normals normally black? I seem to remember them being black, not transparent. The faces in the screenshots are actually transparent / not there. As you can see on the second sshot, some details of the 'other side' of the mesh are visible through the faces at the top.

O and those aren't the only faces missing / inverted: there were some more, I just created two sshots as examples.

fmessmer:
@gavanderhoorn could you please have a look at `Edit Mode` in Blender yourself...all faces are there! 
I'd really appreciate if you could verify your statements before posting them "from memory"...

I don't know why the faces are transparent here and black there...I'm not a graphics expert...I really tried to fix this, but I just don't know the right tools/procedure!
And by the way the "holes" are there in hydro-devel as well...

gavanderhoorn:
>  could you please have a look at Edit Mode in Blender yourself...all faces are there!
> I'd really appreciate if you could verify your statements before posting them "from memory"...

It was an honest question .. nothing more. And I just looked at the meshes in Meshlab, and it seems your right: the faces do have inverted normals.

> I don't know why the faces are transparent here and black there...I'm not a graphics expert...I really tried to fix this, but I just don't know the right tools/procedure!

Ok. Not an expert here either. Trying to use _Re-orient all faces coherently_ errors out with "Mesh has some non 2-manifold faces", so I'm guessing that should be fixed first.

> And by the way the "holes" are there in hydro-devel as well...

Seeing as the title of this PR is "Enhancement/update meshes", I figured this might be something we wanted to address. If not, I'm guessing we can merge this. Let's see if we can address the face-normals in another issue/PR.

fmessmer:
I guess there is no chance to fix this based on these meshes...it's non-manifold edges all over...maybe if we had the originial CAD data...

shoulder_manifoldness
forearm_manifoldness

gavanderhoorn:
As the state of the visual meshes does not influence important functionality (like collision checking), I think we can leave it as a cosmetic issue and fix it later. There are quite some pages explaining how to fixup meshes that are non-manifold ([this](http://www.shapeways.com/tutorials/fixing-non-manifold-models) one from shapeways fi).

> mayb if we had the originial CAD data

<del>Where did @kphawkins create the current meshes from, do we know?</del> -> ur site: [Downloads](http://www.universal-robots.com/GB/Products/Downloads.aspx)

fmessmer:
I also googled a bit on manifoldness-fixing....they all assume you have a "few" non-manifold edges and then remove them...
If we remove all non-manifold-edges we end up in an empty mesh...

gavanderhoorn:
I propose we leave the cosmetic issue out of this. The other changes are already a good improvement.

abubeck:
Ok, based on the discussion I will merge this PR.

PR #92, opened 2014-08-21
IO-Support (fixed and rebased)

fmessmer:
This is a rebased version of PR #81 by @airplanesrule
It includes the fixes from https://github.com/jrgnicho/universal_robot/tree/IO-Fixed provided by @jrgnicho
- some additional fixes and restructuring to be more conform with the ur_driver package layout

It's tested on a UR5 v1.7!
Comments appreciated...


fmessmer:
In case this PR gets accepted #81 is obsolete...(all changes from #81 are included in this PR)!

abubeck:
Any feedback? @airplanesrule : could you please test the updated version on your controller? Or do you also work with v1.7 of the controller.

airplanesrule:
I currently do not have access to a UR5 to test it on, but I should have access to one within a week or two. I will post results here when I regain access to it. If anyone else could test it to make sure it works, that would be greatly appreciated. As for the driver it used, I _think_ it had 2.something on it, but I could be very wrong.

ur_driver/CMakeLists.txt
```diff
@@ -4,86 +4,20 @@ project(ur_driver)
 ## Find catkin macros and libraries
 ## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
 ## is used, also find other catkin packages
-find_package(catkin REQUIRED)
+find_package(catkin REQUIRED COMPONENTS roscpp rospy std_msgs ur_msgs)
```
jrgnicho: I wasn't able to build the until I removed the ur_msgs dependency from the find_package(...) list.  Including ur_msgs here causes catkin to try and build the ur_msgs package twice and errors out.

jrgnicho:
After a small modification to the CMakeLists.txt file in the ur_driver package (see comment above) I tested the driver and it worked well.  However, the ur shakes when no motion is being sent to the controller, not sure if this issue has been reported yet.  I ran it in a Ur controller version 1.6.

shaun-edwards:
@jrgnicho, is this a new issue with this PR, or just an existing problem.

ur_driver/CMakeLists.txt
fmessmer: as there are no build_depends defined in the package.xml there should not be any components listed in find_package anyway....and we do not have build_depends as it is a python package...

fmessmer:
I updated the branch with the latest hydro-devel.
(But even before you should have been able to build the branch....)

As to the "shaking":
For me it is more of an internal vibration and this is known (at least it has been there before).
I don't know where it comes from and what causes it, but this PR is almost certainly not the reason.
Also, the vibration doesn't actually move the arm, it's probably just something with the breaks or so...

@jrgnicho could you please test again and confirm that it now works for you?

jrgnicho:
I was able to build this update successfully.  
I hadn't observe the vibration before but based on the previous comments it is a known problem and should be logged as a separate issue.

fmessmer:
I cannot detect any different behavior between ros-industrial/hydro-devel and the IO-Support branch
So for me, there is no problem....
@jrgnicho please open an issue with a more detailed description!

abubeck:
Since this was tested multiple times and all feedback has been included, I will merge this PR. Thanks for all contributors.

PR #96, opened 2014-08-27
[Hydro] updating moveit_configs

fmessmer:
This PR merges #83 with the current hydro-devel branch and updates the urX_moveit_config packages based on the current models in ur_description


abubeck:
@ipa-fxm has tested this on our UR's. Since there was no additional feedback I will merge this.

PR #99, opened 2014-09-03
[Indigo] IO-Support

fmessmer:


abubeck:
For reference of discussion see #92.

PR #102, opened 2014-09-05
[Hydro] Issue/heavy payload rebased

fmessmer:
This PR supersedes #78 
It is rebased to latest hydro-devel, it adresses the comments from @gavanderhoorn and has some beautifying changes guaranteeing conformity


ur_bringup/launch/ur10_bringup.launch
```diff
@@ -12,6 +12,8 @@
   <arg name="robot_ip"/>
   <arg name="reverse_port" default="50001"/>
   <arg name="limited" default="false"/>
+  <arg name="min_payload"  default="0.0"/>
+  <arg name="max_payload"  default="20.0"/>
```
shaun-edwards, line 16: Isn't the max payload 10kg for the UR 10?  In my experience if you set the payload higher than the rated payload (from the pendant), the robot slowly floats up due to gravity compensation (at least on version 1.4).

fmessmer, line 16: The default value for `max_payload` derives from what the panel allows to be manually set for payload.
I.e. - for the UR10 - if you enter a value higher than 20 on the panel, the panel complains. (For UR5, the value is 10 respectively)

abubeck:
As this was tested on the robot, I will merge this PR.

Issue #108, opened 2014-09-12
driver: Warning: 32 is no a valid flag number


gavanderhoorn:
Running `hydro-devel` @ba8e2270 against `ursim` results in continuous printing of `Warning: 32 is no a valid flag number` (sic).

It would appear that the URScript program tries to access flag 32 (see [here](https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur_driver/prog#L74)). As flags run from [0, 31] (according to the v1.8 scripting manual), that is an invalid index.

It does not appear to influence the rest of the driver, but should be fixed none-the-less.

PR #110, opened 2014-09-16
Fix for issue #108: Warning: 32 is no a valid flag number

gavanderhoorn:
As per subject.


gavanderhoorn:
This is against `hydro-devel` (currently the default branch).

jrgnicho:
+1

abubeck:
Makes sense. I will merge.

Issue #109, opened 2014-09-13
driver: analog IO scaling factor is magic nr


gavanderhoorn:
As per subject:

``` python
socket_send_int(get_analog_out(0)*1000000)
```

A new `MULT_something` should be added.

fmessmer:
@gavanderhoorn for something as minor as that, could you please send a pull request rather than an issue? you can even do the changes in your browser...

gavanderhoorn:
It was a question of time and effort: I just noticed it, but don't have the time right now to fix it. At least it is now a known issue.

Issue #107, opened 2014-09-11
Controller read too many bytes for IO message (and possibly others)


shaun-edwards:
The program that runs on the UR controller reads too many bytes from the socket connection for setting digital outputs.  See [here](https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur_driver/prog#L226).  Most times this is not an issue because multiple commands are not typically in the socket buffer (the reference read command only returns the 2 values used for setting digital IO).  However, when trajectories and IO are driven from the ROS side simultaneously (i.e. from multiple threads), this can occur.  When this does, part of the _next_ message in the buffer is read.  This results in several `unrecognized command` messages being reported in the ROS log (although a more critical failure could be misinterpreting the data as a valid command).

I will submit a PR to address this within the week...but I wanted to document it in the near term.

fmessmer:
Is this related to #106?

gavanderhoorn:
Was this also tested on v1.4? I ask because I've had to use some tricks in the past when reading from sockets on older versions of UR's software.

shaun-edwards:
It was discovered on version 1.4, but is clearly a bug on inspection.  I don't think this is related to #106, since the shaking occurs when no commands are sent to the controller.

gavanderhoorn:
``` python
params_mult = socket_read_binary_integer(1+6+1)
```

Ah, you're right. Seems number-of-bytes was specified, instead of number-of-items.

shaun-edwards:
I should have mentioned, I think some of the other byte reads may also have this problem (possibly copy & paste issue)

gavanderhoorn:
It is probably a copy-paste issue: the same sequence of statements is used in the bodies of the other cases (`MSG_MOVEJ` and `MSG_SERVOJ`). Those seem ok though?

PR #111, opened 2014-09-16
Fix for issue #109: analog IO scaling factor is magic nr

gavanderhoorn:
As per subject.

I have not touched anything else (#107 fi).


gavanderhoorn:
This is also against `hydro-devel`.

jrgnicho:
+1

fmessmer:
+1

abubeck:
Will be applied with #117

Issue #113, opened 2014-09-21
kinematics: errors in build script and manifest.


gavanderhoorn:
According to `catkin_lint -W2 ur_kinematics`:

``` bash
ur_kinematics: error: exported library 'ur5_moveit_plugin' is not installed
ur_kinematics: error: exported library 'ur10_moveit_plugin' is not installed
ur_kinematics: CMakeLists.txt(49): error: catkin_package() lists 'moveit_core' as system package but it is not
ur_kinematics: warning: package 'geometry_msgs' should be listed in catkin_package()
ur_kinematics: CMakeLists.txt(67): warning: use of link_directories() is strongly discouraged
ur_kinematics: CMakeLists.txt(68): warning: use of link_directories() is strongly discouraged
```

PR #114, opened 2014-09-21
Fix for issue #113: issues with ur_kinematics build script and manifest

gavanderhoorn:
As per subject.

I've also added A. Bubeck as an additional maintainer (more in-line with other pkgs in this repository).

Note: `catkin_lint` still complains about `boost` being unconfigured, but this seems to be due to a difference in capitalisation between the CMake Findscript and the rosdep entry.

### DIFF
--- a/ur_kinematics/CMakeLists.txt
+++ b/ur_kinematics/CMakeLists.txt
@@ -1,76 +1,27 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(ur_kinematics)
 
-## Find catkin macros and libraries
-## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
-## is used, also find other catkin packages
-find_package(catkin REQUIRED COMPONENTS roscpp geometry_msgs moveit_core moveit_ros_planning pluginlib tf_conversions)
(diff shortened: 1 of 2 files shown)


fmessmer:
According to [this](http://wiki.ros.org/catkin/CMakeLists.txt), there is no need to specify a dependency to boost in the `package.xml` or `catkin_package(DEPENDS)`.
Simply keeping the `find_package(Boost REQUIRED COMPONENTS system)` solves the `catkin_lint -W2` warning and still compiles successfully.

gavanderhoorn:
@ipa-fxm: could you be a bit more specific? As far as I can tell, Boost is no different from any other system dependency. If you look at `pluginlib` fi ([CMakeLists.txt](https://github.com/ros/pluginlib/blob/indigo-devel/CMakeLists.txt#L5-L14), [package.xml](https://github.com/ros/pluginlib/blob/indigo-devel/package.xml#L20)), it also specifies the Boost dependency in the way I did.

The `catkin_package(.. DEPENDS boost)` bit contains a typo: boost should be capitalised (at least that's what `pluginlib` does).

It probably works without those as Boost is used by core ROS packages. I've always understood it's still advised to explicitly declare the dependency though.

abubeck:
Cleanups are welcome. I will merge this.

PR #115, opened 2014-09-21
Fix for issue #107: Controller read too many bytes for IO message

gavanderhoorn:
As reported by @shaun-edwards: seems like a confusion between _nr-of-items_ vs _nr-of-bytes_ for the `socket_read_binary_integer(..)` statement in `prog`.

PS: I've not had an opportunity to test these changes on actual hardware. I would appreciate it if a reviewer could do that.

### DIFF
--- a/ur_driver/prog
+++ b/ur_driver/prog
@@ -223,41 +223,41 @@ def driverProg():
       elif mtype == MSG_SET_DIGITAL_OUT:
         #send_out("Received Digital Out Signal")
         # Reads the parameters
-        params_mult = socket_read_binary_integer(1+6+1)
+        params_mult = socket_read_binary_integer(2)
         if params_mult[0] == 0:
-          send_out("Received no parameters for set_out message")
+          send_out("Received no parameters for set_digital_out message")
         end
         if params_mult[2] > 0:
-	  set_digital_out(params_mult[1], True)
+           set_digital_out(params_mult[1], True)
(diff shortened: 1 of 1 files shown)


shaun-edwards:
Documented in #107....we/I would verify but our robot is out for repairs.  @jrgnicho, can you verify this vs our local version?

jrgnicho:
Our local version reads two items "params_mult = socket_read_binary_integer(1+1)" when parsing the MSG_SET_DIGITAL_OUT command.  For all the other non-motion commands such as MSG_SET_FLAG, MSG_SET_ANALOG_OUT and  MSG_SET_TOOL_VOLTAGE we are still reading 8  items "params_mult = socket_read_binary_integer(1+6+1)"

gavanderhoorn:
I'm guessing those other _non-motion commands_ are the ones that "[..] may also have this problem (possibly copy & paste issue)" (@shaun-edwards in [107](https://github.com/ros-industrial/universal_robot/issues/107#issuecomment-55397466)).

All those messages use the same format and have the same length though: `msg_id, pin, val`. Only for `SET_TOOL_VOLTAGE` is `pin` used as `val`.

gavanderhoorn:
There are more issues with the current IO implementation, but I'm guessing this can at least be merged?

abubeck:
Change will be applied with #117

Issue #36, opened 2014-02-05
Remove BeautifulSoup dependency


mikepurvis:
It's not available in Ubuntu until Raring—that aside, it's way overkill for just parsing urdf. Given the [minimal functionality being employed](https://github.com/ros-industrial/universal_robot/search?q=BeautifulSoup), almost anything would work, but [ElementTree](http://docs.python.org/2/library/xml.etree.elementtree.html) has been standard since Python 2.5.

The ideal solution would be if the urdf package provided a Python parser, but [that is not the case](http://wiki.ros.org/urdf#Overview). Taking this step would make universal_robot releasable into Hydro.

shaun-edwards:
@mikepurvis - Thanks for bringing this up.  I agree the BeautifulSoup dependency is causing issues.  We plan to release universal into hydro very soon, so this is something we will have to address.  The URDF information is used to load joint offsets.  When this fails (because we don't have them in the URDF), a warning is posted to the log.  This has cause confusion several times.  I'm considering pulling out this entirely.  Nobody really knows if or why the joint offsets are required.  I don't know of anybody that is actually using them and they were never included in an actual release.  Long story short, it's a bad dependency, it supports a feature(unreleased) nobody uses, and there doesn't seem to be a good reason to keep it.

FYI, there are some python urdf parsers out there, [urdf_parser_py](http://wiki.ros.org/urdf_parser_py), but since they aren't actively supported we cannot depend on them.

mikepurvis:
Given how BeautifulSoup is being used, though, just a basic XML parser should be more than enough. Or, as you say, dump the whole feature.

We (Clearpath) are pretty interested in this driver— can you provide a more specific timeline for your planned release into Hydro?

kphawkins:
https://github.com/ros/urdfdom
urdfdom is the stack where all of that development ended up. It's
maintained fairly well and will eventually be incorporated into ros core,
I'm sure. I'm pretty sure it has a .deb now. They managed to get rid of
that dependency by writing or incorporating a new xml parser which is
included, I think.

shaun-edwards:
@mikepurvis - we are in the process of performing release right now.  The next package to be released will be abb, followed by universal.  I expect this to happen before the end of February.

mikepurvis:
Sweet, thanks!

PR #116, opened 2014-09-21
Fix for issue #36: Remove BeautifulSoup dependency

gavanderhoorn:
As per subject.

Run-time behaviour should be identical. Tested with ur-sim.

Note: I opted to use lxml instead of `urdfdom_py`, as the latter has some issues parsing the UR5/10 urdfs and we only need access to a single custom element. Rosdep keys for `python-lxml` were recently added (for `urdfdom`) for all (Ubuntu) versions, so this should not be an issue.

Error handling is rather primitive right now (as in: user is immediately presented with stacktraces), but I did not want to change that in this PR.

### DIFF
--- a/ur_driver/package.xml
+++ b/ur_driver/package.xml
@@ -12,18 +12,13 @@
   <author>Stuart Glaser</author>
   <author email="sedwards@swri.org">Shaun Edwards</author>
 
-
   <buildtool_depend>catkin</buildtool_depend>
-  
+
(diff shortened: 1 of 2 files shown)


fmessmer:
Looks all right...tested with UR5 v1.7...no behavioral change noticed....

gavanderhoorn:
So +1?

abubeck:
As it seems to work, I will merge.

PR #121, opened 2014-10-09
Decimated and converted to STL ur5 collision meshes.

hersh:
This is a workaround for a bug in moveit which I haven't solved yet: https://github.com/ros-planning/moveit_core/issues/202

In any case, these models much more closely match the geometry of the robot since they are not convex.  I did "decimate" them in Blender to reduce the polygon count to make collision checking faster though.


fmessmer:
Thanks @hersh for digging into this.
Those sound like two major bugs for MoveIt! 

Personally, I'm fine with using stl's for collision (as using the collada for visual seem to be ok).
Personally, I'm also fine with using the same (but reduced) visual meshes for collision, too. 
However, there is this ROS-Industrial [tutorial](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Collision_STLs) that suggests to use convex hulls.
So - yet again - this can be discussed.

One more thing:
For consistency, would you mind doing the same changes for the UR10 as well? I.e. generate the collision meshes?

Oh, and do the new collision meshes have any effect on the self-collision matrix in the srdf file? Did you re-run the self-collision checking in the Setup Assistant?

tnl-woi:
Thanks for this!
I asked a [question](http://answers.ros.org/question/194600/ur5-self-collisions/) about this yesterday, because my UR5 was often colliding with itself.

I applied your changes (stl-collision-meshes and ee-link) to my system and the robot didn't have any collisions with itself anymore, so this seems to solve (or work-around) the problem.

As I already mentioned in the question, I think the self-collision matrix is wrong, or my understanding of it is incorrect. I assume the srdf file just describes which links cannot physically collide with each other, which can quite easily be checked by visual inspection. If this is the case, I think the following lines in the srdf are incorrect and should be removed:

`<disable_collisions link1="forearm_link" link2="shoulder_link" reason="Never" />`
`<disable_collisions link1="forearm_link" link2="wrist_2_link" reason="Never" />`
`<disable_collisions link1="forearm_link" link2="wrist_3_link" reason="Never" />`
`<disable_collisions link1="upper_arm_link" link2="wrist_1_link" reason="Never" />`

I also think wrist_1_link cannot collide with the shoulder_link, but a second opinion on that would be great ;-)

hersh:
I just ran moveit_setup_assistant's self-collision checker on the original high-res but convex mesh version of the UR5 and my new low-res concave mesh version and timed it (with a stopwatch).  I ran 3 tries for each, on the highest setting (100,000 samples):

original version:
8.53
8.39
8.36

my new version:
2.56
2.44
2.42

Units are seconds.  So it seems clear that even if it is logically better to use only convex meshes, there is a bigger improvement from using simplified meshes, even if they are not convex.

To @tnl-woi's point about the 4 disable-collisions lines in srdf: they are indeed wrong.  The bug https://github.com/ros-planning/moveit_core/issues/202 affects moveit_setup_assistant and its self-collision checker as well as the planning code.  When the self-collision checker was originally run to create the moveit_config package, it incorrectly found that the forearm_link never collided with the shoulder_link, etc.

tnl-woi:
Hi @hersh,
I noticed in rivz that the collision model is always red now, while the previous model was white. Is this just a color change, or does this mean the system things it's in colision?
This change was observed after cherry-picking this specific pull request in my indigo branch. Is this the expected result?

hersh:
@tnl-woi, I doubt the color was changed intentionally, it probably means it thinks there is a collision.  The moveit_setup_assistant collision-checker probably needs to be re-run with the new models.

fmessmer:
In #128 I build on top of this PR applying the changes for the UR10 as well + re-generate URDFs + re-run moveit-setup-assistant

abubeck:
This is tested and will be extended by #128 
I will merge this now.

PR #122, opened 2014-10-09
Workaround for moveit collision-checking bug

hersh:
This adds a tiny cube to the collision part of the ee_link of the ur5, and also adds it to the .srdf in the moveit_config package so collisions between it and wrist_3 get ignored.  It is a workaround for https://github.com/ros-planning/moveit_core/issues/158.  The way the urdf and srdf are arranged, it is common to attach grasped objects to the ee_link, at which point they will be ignored by the collision-checking code in moveit_core.

This pull request is chained off of https://github.com/ros-industrial/universal_robot/pull/121 for my own convenience.  If you don't want that one but you do want this one, I can cherry-pick it and resubmit.


fmessmer:
I think that tiny box is not a problem if it solves collision checking with attached objects.
If it's just a matter of **having** a geometry, we could check to chose a geometry that does not extrude the writst_3_link mesh, i.e. ee_link geometry to be _within_ the wrist_3_link geometry - of course not changing the ee_link frame but the orrigin of the collision tag. What do you think?

I think it's ok to not have a visual. I'd just like to mention this to get confirmation.

And, again, could you please also do the changes for the UR10?

hersh:
I put the X coordinate of the 1cm box at -1cm, so the ee_link geometry is embedded within the wrist_3_link geometry.  Looking at the model, it blocks the center hole in wrist_3_link, but does not extend beyond the outer mounting surface.  It could easily be pushed further in (-2cm?) so as to be completely invisible.

gavanderhoorn:
@hersh: if I understand you (and the MoveIt bug) correctly, this is something that would affect all ROS-Industrial MoveIt configuration packages made for manipulators that have a `ee_link` or `tool0` link without any geometry defined for them?

If so, we should make this a ROS-I global issue over at [ros-industrial/ros_industrial_issues](https://github.com/ros-industrial/ros_industrial_issues).

hersh:
@gavanderhoorn I said it was common to attach objects to the last link in the arm group, but I'm not really sure how true that is.  It is what I've been doing in my custom code, but it's certainly possible that other code attaches objects to gripper links etc.  However it is so convenient to use virtual links that I'll bet this affects more than just the Universal robots.

fmessmer:
In #128 I build on top of this PR applying the changes for the UR10 as well + re-generate URDFs + re-run moveit-setup-assistant

abubeck:
This was tested and will be extended by #128.
I will merge this now.

psfafz:
can you please also merge it for the ingigo version ?

tnl-woi:
@psfafz see #130

PR #128, opened 2014-10-21
Moveit issues hydro

fmessmer:
This PR builds on top of #121 and #122 
It adds new collision meshes for the UR10 (simplified stl's). Number of vertices for UR10 collision meshes are about the same as for UR5 collision meshes.
It also re-generates the urX_moveit_configs (i.e. self-collision-matrix!)
It makes the KDLKinematicsPlugin the default IK-Plugin again, as the ur_kinematics plugins fail for joint_limited versions of UR5 and UR10 respectively.
It also re-generates the URDF files from the latest XACRO files

PR #130, opened 2014-10-22
[Indigo] Bring hydro updates to indigo

fmessmer:
This rebases the latest developments from `hydro-devel` to `indigo-devel`

Please test!


tnl-woi:
:+1: I've tested these changes last week on the UR5 CB3 and it worked fine for me, without additional changes to the universal_robot package (which were necessary before to get it working). I was able to successfully move the robot without (self-)collissions using MoveIt!.
Note, however, that the collision model in rviz is still red and I'm not sure if this indicates a problem in the code or the (updated) collision models or if it's something else...

fmessmer:
> Note, however, that the collision model in rviz is still red and I'm not sure if this indicates a problem in the code or the (updated) collision models or if it's something else...

I think the collision meshes now are red because we now use stl files and not collada anymore.
stl file do not have a color/texture information in the file format - while collada files do.
Thus, I think red is just the default color used for collision meshes in case no color/texture information can be found within the mesh file itself.

I tested this by using very small boxes `<box size="0.01 0.01 0.01"/>` for all the links instead of the meshes. Thus, I can be sure they do not collide. I also commented all `<disable_collision>` entries in the moveit_config. Still they are red.
When I create similar boxes in blender, assign a color, export as collada and use those collada meshes for collision, they are displayed in the color/texture I assigned in Blender...

Unfortunately, the `<material>` in the URDF is only considered for visual and not for collision. Thus, I guess we cannot change this...

gavanderhoorn:
> I think the collision meshes now are red because we now use stl files and not collada anymore.
> stl file do not have a color/texture information in the file format - while collada files do.
> Thus, I think red is just the default color used for collision meshes in case no color/texture information can be found within the mesh file itself.

For `fanuc` and `fanuc_experimental` we only use stls, and I have not observed this behaviour. Only the `visual` elements have `material` children, and both the collision and the visual geometry shows the assigned colours when launching any `demo.launch` from any of the moveit cfg pkgs.

One difference is that the Fanuc models are not using collada anywhere, so perhaps this is a bug that only shows when using different mesh formats for visual and collision geometry.

> I also commented all `<disable_collision>` entries in the moveit_config. Still they are red.

Wouldn't doing that just cause all links to be considered in collision detection? Not only those that are not normally already 'in collision' (because they are adjacent fi).

fmessmer:
@gavanderhoorn you are right.
Adding

```
        <material name="Cyan">
          <color rgba="0 1.0 1.0 1.0"/>
        </material>
```

to the visual tag of the links in the URDF, will color the collision meshes in rviz Cyan as well.....
This color is used both within the RobotModel Display as well as the MotionPlanning Display (for SceneRobot and PlannedPath) when "Show Robot Collision" is enabled.
The visual meshes and Gazebo ar not affected.

@gavanderhoorn @tnl-woi 
Thus, what color do we want?

gavanderhoorn:
I don't really have a preference for a colour.

For `fanuc`, I've kept all collision geometry colours the same as the visual ones.

Seeing as `universal_robot` uses collada meshes, perhaps a colour that approximates the overall colour of the visual mesh could be used? It should already be clear from the cruder meshes that one is looking at the collision meshes.

fmessmer:
I chose a LightGrey color 

```
        <material name="LightGrey">
          <color rgba="0.7 0.7 0.7 1.0"/>
        </material>
```

I applied the according changes in #128.

abubeck:
Main fixes (collada and urdf updates) are included and have been reviewed. Color updates will be rebased to indigo in a following PR. They are already applied in #128.

psfafz:
I encounter an error (failure of reading the current joint states) when brining the ur5 up after the update. Is this the right platform to discuss?

abubeck:
@psfafz: yes it is :) could you please give us more information on the version of the UR control software you are using? Thanks.

psfafz:
From the control panel of the robot I read "URControl 1.5.7849 (Feb 06 2012)".

When brining the robot up I get:
........
[INFO] [WallTime: 1415093843.261606] Max velocity accepted by ur_driver: 10.0 [rad/s]
[INFO] [WallTime: 1415093843.264319] Bounds for Payload: [0.0, 10.0]
Disconnected.  Reconnecting
[INFO] [WallTime: 1415093843.312926] Programming the robot
[INFO] [WallTime: 1415093843.314152] Programming the robot at 192.168.1.1
RobotStateRT has wrong length: 756
Exception in thread URConnectionRT:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(_self.__args, *_self.__kwargs)
  File "/home/faps/FZ_WS/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 374, in __run
    self.__on_packet(packet)
  File "/home/faps/FZ_WS/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 340, in __on_packet
    for i, q in enumerate(stateRT.q_actual):
AttributeError: 'RobotStateRT' object has no attribute 'q_actual'

Handling a request
[INFO] [WallTime: 1415093843.914763] Robot connected
[2014-11-04 10:37:23.916431] Out: hello

When launching moveit and rviz the displayed robot configuration does not match the physical configuration

abubeck:
You have version 1.5 of URControl. The current version of the ROS driver (and the versions before) have only been tested with URControl version 1.6, 1.7, 1.8 and 3.0. UR has no protocol information for controllers <1.6 anymore on the website. If you can provide us with the protocol specification for version 1.5 we can try to also support this old version of the controller. Otherwise I recommend upgrading to a newer version.

tnl-woi:
> Seeing as universal_robot uses collada meshes, perhaps a colour that approximates the overall colour of the visual mesh could be used? It should already be clear from the cruder meshes that one is looking at the collision meshes.

I did not observe that the collision meshes were cruder in this particular version (maybe it was less noticeable due to the color?), but it seems all right now (tested with tag 1.1.5). Color is fine and collision meshes are distinguishable from the visual meshes. :+1: Thanks all!

gavanderhoorn:
@abubeck: the older versions of the driver actually worked with 1.4.x, and I think SwRI has / had a controller running version 1.5.x of Polyscope.

> If you can provide us with the protocol specification for version 1.5 we can try to also support this old version of the controller.

[Rev4](http://wiki03.lynero.net/Technical/RealTimeClientInterface?rev=4) of the _RealTimeClientInterface_ wiki page would seem to document a packet with a length of 756 bytes.

psfafz:
I think I will update my controler and try again-
update:
I just found out that this UR5 is so old that 1.5 is the lastest possible controler version. How can I give you the protocol specification for 1.5 ?

gavanderhoorn:
@psfafz: afaik all CB2 controllers can be updated to the 1.8 version. Do you have a CB1?

Also: in order to upgrade from 1.5, you have to first upgrade to 1.6, then 1.7, etc.

psfafz:
I dont know which one I have - the production date was 1/2010

gavanderhoorn:
@psfafz wrote:

> I just found out that this UR5 is so old that 1.5 is the lastest possible controler version

How did you determine this?

psfafz:
That was information from my advisor. He suggested to back to groovy since he had this robot running with it quite stable

gavanderhoorn:
Personally I wouldn't want to revert to groovy, if only for the fact that MoveIt has advanced quite a bit in Hydro.

Whether you want to update or not, I'll leave to you.

Just in case: the 1.4 version expects an `info.xml` file next to the `.tgz` file that contains the actual update. Without that first file, it will afaik not want to update. You _have_ to update to 1.5 first. 1.6+ don't have the `info.xml` file anymore.

Easiest is to go to [ur-update.dk/](http://ur-update.dk/URsupport/Software/Robot_software/) and download the appropriate files onto usb stick. Update procedure itself is not too difficult.

jeppewalther:
@psfafz Did you come up with a solution

psfafz:
@jeppewalther  I did go back to groovy. Since I am mostly just a user I am satisfied with this version since everything works as expected (no more self-collisions nor path planning problems). However, I do not know what advantages the current version has to offer.

Issue #118, opened 2014-09-30
Revision of IO-Interface


fmessmer:
The current IO-interface has been discussed in #79 and is available since #92 
However, there are some shortcomings. A revision is suggested!

As already mentioned in #106, the IO-Support needs a significant revision in order to reflect e.g. analog_input/output_range/domain as can be seen in the MasterboardData structure of deserialize.py

gavanderhoorn:
To do this properly we need to implement setters for the ranges (extension of `prog`), and take the currently configured ranges into account when converting reported analogue in / out values.

@ipa-fxm: is that a 'significant revision'? :)

fmessmer:
Yes, that's what I mean by "significant revision".
The current "IO-Interface" simply does not provide a generic IO-Interface. There is a lack of information in the current `/io-states` message (e.g. range). As well as it makes assumptions, e.g. only allows to set Voltage for analog_in/out.

Also error handling might have a potential for improvement...(see this [comment](https://github.com/ros-industrial/universal_robot/pull/117#discussion_r18220875))

fmessmer:
We should also investigate @shaun-edwards 's [comment](https://github.com/ros-industrial/universal_robot/pull/117#discussion_r18220636)

gavanderhoorn:
I don't think range should be included in the io state topic. It is only needed in the conversion to the actual value sensed by the adc, so the driver should have access to it. I'd provide getters as services to interested users, but it does not need to be included in the topic itself (as it is more of a configuration item).

fmessmer:
On more comment form @tnl-woi about IO-Interface with v3.0 that needs to be investigated:

> I checked the /io_states topic. It is published at (almost) 20 Hz.
> 
> I'm pretty sure the digital inputs and outputs are swapped: the digital_in_states properly reflect what I enter on the teach pendant, while the digital_out_states are quite noisy (i.e. toggling).
> Pins 8 and 9 are never updated. According to the documentation get_digital_in and get_digital_out are depecrated. There are now separate functions for pins 0-7 and for pins 8-9. The old function is still supported for backwards compatibility. But seeing this behavior I doubt this is the case.
> 
> The analog input and output values are always 0.0, even if I change the output using the TP. But I don't have something connected to the outputs, so I'm not sure what to expect here...

tnl-woi:
[...]

PR #135, opened 2014-10-28
version-aware implementation for deserialize and deserializeRT in order ...

### LINKED ISSUE #123: Strategy for handling different data structure formats for different controller versions
See [comment](https://github.com/ros-industrial/universal_robot/pull/117#issuecomment-58496562)

fmessmer:
...to support v3.0

fixes #123 

---

Currently tested with:
**UR5**
- [x] 1.6
- [x] 1.7
- [x] 1.8
- [x] 3.0

**UR10**
- [x] 1.6
- [x] 1.7
- [x] 1.8
- [ ] 3.0

---

**Note**
This PR does not include the fixes for the MoveIt! configs!

### DIFF
--- a/ur_driver/src/ur_driver/deserialize.py
+++ b/ur_driver/src/ur_driver/deserialize.py
@@ -50,7 +50,7 @@ class ToolMode(object):
     BOOTLOADER = 249
     RUNNING = 253
     IDLE = 255
-    
+
 class MasterSafetyState(object):
     UNDEFINED = 0
     BOOTLOADER = 1
@@ -76,7 +76,7 @@ class RobotModeData(object):
         (plen, ptype) = struct.unpack_from("!IB", buf)
         if plen == 29:
             return RobotModeData_V18.unpack(buf)
-        elif plen == 37:
+        elif plen == 38:
             return RobotModeData_V30.unpack(buf)
         else:
             print "RobotModeData has wrong length: " + str(plen)
@@ -103,15 +103,16 @@ class RobotModeData_V30(object):
     __slots__ = ['timestamp', 'robot_connected', 'real_robot_enabled',
                  'power_on_robot', 'emergency_stopped',
                  'security_stopped', 'program_running', 'program_paused',
-                 'robot_mode', 'target_speed_fraction', 'speed_scaling']
+                 'robot_mode', 'control_mode', 'target_speed_fraction',
+                 'speed_scaling']
     @staticmethod
     def unpack(buf):
         rmd = RobotModeData_V30()
         (_, _,
          rmd.timestamp, rmd.robot_connected, rmd.real_robot_enabled,
          rmd.power_on_robot, rmd.emergency_stopped, rmd.security_stopped,
-         rmd.program_running, rmd.program_paused, rmd.robot_mode,
-         rmd.target_speed_fraction, rmd.speed_scaling) = struct.unpack_from("!IBQ???????Bdd", buf)
+         rmd.program_running, rmd.program_paused, rmd.robot_mode, rmd.control_mode,
+         rmd.target_speed_fraction, rmd.speed_scaling) = struct.unpack_from("!IBQ???????BBdd", buf)
         return rmd
 
 #this parses JointData for all versions (i.e. 1.6, 1.7, 1.8, 3.0)
@@ -125,7 +126,7 @@ class JointData(object):
         for i in range(6):
             jd = JointData()
             (jd.q_actual, jd.q_target, jd.qd_actual, jd.I_actual, jd.V_actual,
-             jd.T_motor, jd.T_micro, 
+             jd.T_motor, jd.T_micro,
              jd.joint_mode) = struct.unpack_from("!dddffffB", buf, offset)
             offset += 41
             all_joints.append(jd)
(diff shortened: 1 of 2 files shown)


fmessmer:
With this PR all Polyscope Versions should be supported, i.e. v1.6, v1.7, v1.8 and v3.0

I tested it on a UR5 with controller version **1.8.11997**
Unfortunately, we don't have any other hardware around at the moment!

Please test this PR and provide feedback!
Please also state the controller version you tested!

gavanderhoorn:
Nice.

Btw: for version-differentiated testing, I always like to use [ursim](http://ur-update.dk/URsupport/Software/URSim/). The archives under _installation files_ extract to a ready-to-run simulation of the respective Polyscope version. I'm not sure how good the simulation is, but at least for urscript and rt client testing it seems to work well enough.

fmessmer:
As the changes are really prone to unintentional mistakes, I'd really appriate this PR being carefully reviewed by several pairs-of-eyes... :smiley:

tnl-woi:
I did a quick test with UR5 v3.0. I have to do some more testing, but one thing I can mention already is that I consistently get the following Exception during startup:

```
setting /run_id to bd850b62-5eba-11e4-acdb-001e4fc8ca8d
process[rosout-1]: started with pid [6234]
started core service [/rosout]
process[robot_state_publisher-2]: started with pid [6251]
process[ur_driver-3]: started with pid [6274]
process[tf2_buffer_server-4]: started with pid [6278]
Setting prefix to 
[WARN] [WallTime: 1414511688.414971] No calibration offset for joint "shoulder_pan_joint"
[WARN] [WallTime: 1414511688.415279] No calibration offset for joint "shoulder_lift_joint"
[WARN] [WallTime: 1414511688.415472] No calibration offset for joint "elbow_joint"
[WARN] [WallTime: 1414511688.415665] No calibration offset for joint "wrist_1_joint"
[WARN] [WallTime: 1414511688.415854] No calibration offset for joint "wrist_2_joint"
[WARN] [WallTime: 1414511688.416044] No calibration offset for joint "wrist_3_joint"
[INFO] [WallTime: 1414511688.416244] No calibration offsets loaded from urdf
[INFO] [WallTime: 1414511688.418834] Bounds for Payload: [0.0, 10.0]
Disconnected.  Reconnecting
[INFO] [WallTime: 1414511688.479327] Programming the robot
[INFO] [WallTime: 1414511688.480450] Programming the robot at 10.0.0.9
RobotModeData has wrong length: 38
MasterboardData has wrong length: 72
Euromap67Interface is ignored
Exception in thread URConnection:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/wouter/catkin_ws/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 272, in __run
    self.__on_packet(packet)
  File "/home/wouter/catkin_ws/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 190, in __on_packet
    if not state.robot_mode_data.real_robot_enabled:
AttributeError: 'RobotModeData' object has no attribute 'real_robot_enabled'

Handling a request
[2014-10-28 16:54:49.201420] Out: hello
[INFO] [WallTime: 1414511689.245220] Robot connected
The action server for this driver has been started
```

The robot can still be controlled after this exception. My guess is that the first received packet is incorrect (I remember when I wrote my first program connecting to the UR5 that I had to flush all packets during startup, maybe that's necessary here too).

fmessmer:
@tnl-woi thanks for testing!
Let me know in case you found the cause for the problem...preferably with an according solution/fix...

tnl-woi:
I've found the cause(s) and implemented a fix (see ipa-fxm#5).
NB: Apparently the Python exception caused the thread to be killed, so the URConnection did not process any packets at all.

The cause for the wrong length of the RobotModeData was a forgotten field. I've added this.

The MasterboardData seemed implemented correctly, but apparently the actual packet format differs from the documentation. There are 4 bytes inserted after the header and before all other fields. The value was 0 in my case. I've added a dummy variable to skip this undocumented field, and this solves the exceptions.

fmessmer:
Thanks for the fixes!
How did you find out about the additional byte in the MasterboardData?
Are you sure this is a "meaningless" byte?
Is the `/io_states` topic still publishing the correct IO-States for v3.0?
Can you test this with `rostopic echo /io_states` and then change the IOs via the pendant?

tnl-woi:
The additional bytes are explained in ipa-fxm#5.

I checked the `/io_states` topic. It is published at (almost) 20 Hz.

I'm pretty sure the digital inputs and outputs are swapped: the digital_in_states properly reflect what I enter on the teach pendant, while the digital_out_states are quite noisy (i.e. toggling).
Pins 8 and 9 are never updated. According to the documentation `get_digital_in` and `get_digital_out` are depecrated. There are now separate functions for pins 0-7 and for pins 8-9. The old function is still supported for backwards compatibility. But seeing this behavior I doubt this is the case.

The analog input and output values are always 0.0, even if I change the output using the TP. But I don't have something connected to the outputs, so I'm not sure what to expect here...

gavanderhoorn:
@tnl-woi: could you perhaps make a wireshark capture of some traffic between your CB3 and your ROS pc? You don't need to be running `ur_driver`, a simple `nc ROBOT_IP 30002 > /dev/null` and `nc ROBOT_IP 30003 > /dev/null` (preferably not in the same capture) would be sufficient.

Due to the data rate, you might have to use text mode (`tshark.exe`). Or use `nc` itself, obviously.

fmessmer:
I suggest to focus on the `/io_states` parts when dealing with #118 
As I'm starting to get swamped with branches again...

I added @tnl-woi's comment to the respective issue!

abubeck:
Checked 1.6 and 1.7 as they are conform with 1.8 and have been tested before.
3.0 is tested in UR5 and in in simulation. Therefore I think the tests as they have been done are sufficient. 
I will merge this into the hydro_devel tree.
Thanks to all the testers.

tnl-woi:
I've added an update about the I/O part in #118 and will continue that discussion there, as suggested.

@gavanderhoorn I've sent you the captures by e-mail.

Issue #124, opened 2014-10-10
driver: undocumented max_velocity parameter causes plan execution to fail


gavanderhoorn:
As reported on ROS Answers in [path planning exceeding velocity limits of ur5](http://answers.ros.org/question/194667/path-planning-exceeding-velocity-limits-of-ur5/). Planning with MoveIt often results in:

```
[ERROR] [WallTime: 1412925263.358047] Received a goal with velocities that are higher than 2.000000
```

The private parameter `max_velocity` is currently undocumented (as are all other parameters).

The wiki page for `ur_driver` should probably be extended with an overview of the defined parameters and their values.

fmessmer:
Well MoveIt! plans with the velocity_limits given in [joint_limits.yaml](https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur5_moveit_config/config/joint_limits.yaml)
And those values are retrieved from the velocity limit tags in the URDF description of the joints (see e.g. [here](https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur_description/urdf/ur5.urdf.xacro#L105)).

With a `max_velocity` parameter in the driver having a value less than the MoveIt! velocity limits, this would most likely cause most trajectories to fail.

I thus would increase the default `max_velocity` parameter in the driver to (the maximum) of the MoveIt velocity limits (or have velocity limits per joint in the driver?) in order to allow executing the planned trajectories.

Everyone who wants to move at a slower speed would need to reduce both `max_velocity` parameter in the driver and the MoveIt! velocity limits accordingly.
In order to facilitate this, we could add the private parameter to the ur_common.launch file as suggested by @gavanderhoorn 

Any comments?

shaun-edwards:
I have safely increased the max velocity parameter on my system.  I believe the outlined approach is the right one to take.  It will lead to "safety stops" by the universal because the current driver isn't capable of moving much faster than the `max_velocity` reliably.

gavanderhoorn:
On 30-10-2014 4:19, Shaun Edwards wrote:

> I have safely increased the max velocity parameter on my system.  I believe the outlined approach is the right one to take.  It will lead to "safety stops" by the universal because the current driver isn't capable of moving much faster than the `max_velocity` reliably.  

I think the reason I reported this was not so much the technical issue
of having a parameter that limits the max joint speed the driver is
'comfortable' allowing in a trajectory, but more the fact that that
parameter is completely undocumented, and (worse) has a default that is
[...]

PR #136, opened 2014-10-31
[Hydro] Default max velocity parameter

fmessmer:
as discussed in #124


abubeck:
Discussions are in #124. Solves the feature proposed in discussion. I will merge.

PR #141, opened 2014-11-05
[Indigo] Fix for unpacking masterboard data with UR firmware v3.0.

tnl-woi:
In 1.1.5 support for UR firmware v3.0 was added, but the alignment of the masterboard data was incorrect. This fixes the alignment.
As mentioned in #118.

### DIFF
--- a/ur_driver/src/ur_driver/deserialize.py
+++ b/ur_driver/src/ur_driver/deserialize.py
@@ -203,7 +203,7 @@ class MasterboardData_V30(object):
     @staticmethod
     def unpack(buf):
         md = MasterboardData_V30()
-        (_, _, _UNDOCUMENTED_,
+        (_, _,
          md.digital_input_bits, md.digital_output_bits,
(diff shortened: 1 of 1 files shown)


fmessmer:
LGTM...so in the end it's back to the way it is specified in the wiki [v3.0](http://wiki03.lynero.net/Technical/PrimaryAndSecondaryClientInterface?foswiki_redirect_cache=01011cdbde11a87ca580c01fad780590)?

tnl-woi:
Well, there are still 4 new bytes at the end of the packet, that are not documented (packet size is 72, documentation accounts for 68 bytes). They are currently not unpacked in the code, since we don't know what their meaning is and it's not necessary to unpack more than we actually use.

gavanderhoorn:
Bump (and also #142).

fmessmer:
LGTM!

abubeck:
LGTM, As this has been tested on v3.0 I will merge.

PR #151, opened 2014-12-24
Changed variable params_mult to unique name in each socket_read instance...

### LINKED ISSUE #149: Cannot subsequently execute joint motion and set io through driver
When
- Any **set io** command is sent to controller after a commanded **joint motion**, or
- Any **joint motion** is sent to controller after a **set io** command,

the pendant shows a fault reading something like "Resizing lists not allowed" and the driver on the PC reports something about "EOF received in response" and loses communication with the controller. Driver must be restarted in order to resume operation.

Tested with ROS Hydro on Ubuntu 12.04 with both _hydro-devel_ and _indigo-devel_ branches on a V3 UR controller.

dpsolomon:
Fixes #149. See detailed description of fix in PR #150

### DIFF
--- a/ur_driver/prog
+++ b/ur_driver/prog
@@ -121,23 +121,23 @@ def driverProg():
         break
       elif mtype == MSG_MOVEJ:
         send_out("Received movej")
-        params_mult = socket_read_binary_integer(1+6+4)
-        if params_mult[0] == 0:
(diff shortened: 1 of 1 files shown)


fmessmer:
LGTM!

abubeck:
This patch is important. I just tested this successful on the robot. I will merge this.

PR #156, opened 2015-01-29
Improve moveit rviz plugin params

fmessmer:
This significantly reduces planning time and the planned paths are much smoother/shorter for planning requests triggered from the MotionPlanning plugin in rviz.

These changes are motivated by the discussion on the moveit mailing list [here](https://groups.google.com/forum/#!topic/moveit-users/M71T-GaUNgg). There seems to be a problem within MoveIt! related to parallel planning.


abubeck:
This looks good and only changes the default config (minor change). I will except this.

Issue #87, opened 2014-08-06
Modeling UR5/10's dynamic behavior


arennuit:
Hi all,

The behaviour of the simulated UR5/10 does not replicate the behaviour or the real robots. This is due to 3 problems at least:
1. The parameters of the simulation (geometric and dynamic) do not replicate the real robot's params
   - e.g. the dynamic parameters of the ur5/10 which are currently used for the simulation (in files _ur5.urdf.xacro_ and _ur10.urdf.xacro_) do not mimic the real robots' parameters. They were found making simple assumptions of the robot (i.e. the inertia are computed as the inertia of homogeneous cylinders...)
2. The low-level control of the simulation does not replicate the low-level control of real robots
   - The simulation uses PID based on ros_control, but it is not sure what low-level control method is used by the real robots
3. The simulation may not fully reproduce the behaviour of the real world's physics
   - This is especially true as the physics engines used by gazebo (ODE, bullet...) are RT engines not aimed at accuracy (they trade accuracy off for speed of computation)
   - This question relates to [this post](http://answers.ros.org/question/188953/real-vs-simu-params-and-gains-behavior-difference/?answer=189145#post-id-189145) in ROS answers

This is a complex problem which outreached the boundaries of the UR world, and we cannot expect to solve it all here. Though we can certainly improve the situation and this is what this post is aimed at.

Kind regards,

Antoine.

arennuit:
Problem 1 (parameters) can be improved by performing the identification of a real robot (mainly dynamic as geometric parameters are known from the _urcontrol_ file).

For problem 2, does anyone have an idea of the low-level architecture used in real UR robots?

Problem 3 can be tackled by the choice of the physics engine. In [PR #84](https://github.com/ros-industrial/universal_robot/pull/84#issuecomment-51320245), @ipa-fxm mentioned that gazebo-ODE could not reliably mimic the real world's behaviour because of its inability to simulate friction. I am not sure this is true [(see here why)](http://gazebosim.org/wiki/Tutorials/1.9/friction) but it highlights that physics engine are precision limits. Maybe the choice of one engine or another could influence the overall simulation quality.
- Another idea would be to identify the robot's dynamic parameters not in the real world but in the simulated world to compensate for the physics engine error

Anyone willing to comment on this?

gavanderhoorn:
[...]

PR #161, opened 2015-02-11
[Indigo] Use PositionJointInterface for simulation

fmessmer:
So far, the gazebo simulation was using the EffortJointInterface within the hardwareInterface tag of the transmission URDFs. (This is probably a copy-and-paste-remnant from PR2-times - as PR2-hardware was effort-controlled). This PullRequest changes the hardwareInterface to a PositionJointInterface to improve and simplify the gazebo simulation. 

There are several reasons that motivate this change.
First of all, the actual universal_robot hardware is **not** effort-controlled but accepts position-setpoints!
So using a PositionJointInterface for simulation as well seems reasonable.
The main advantage of using the PositionJointInterface over a EffortJointInterface is the following:
With the PositionJointInterface the JointTrajectoryController actually becomes a straight position_controllers/JointTrajectoryController which then actually is no more than a kind of forward command controller throughpassing the position setpoints from the FollowJointTrajectory goal to position setpoints for the simulated joints in Gazebo.
As the mapping is now position - position and no longer position - effort, the internal PID within the FollowJointTrajectory controller is no longer needed and the parameter tuning of the PID gains within the controller configuraiton yaml is obsolete!
As this particular issue has been discussed several times already, I assume getting rid of it to be a good thing! 
This PR is also related to #87 in a way that no PID "guessing" is required anymore.
It's stil not the same behavior as the real hardware uses a different joint trajectory interpolation and also the low-level aspects of the UR controller are not considered yet...

With respect to interfaces this changes nothing within the ROS runtime system
Gazebo will behave as an ideal controller setting the simulated joints to the position setpoints. Thus, there is no "jerky" motion or oscillation anymore!

This PullRequest has absolutely no effects on running it with the real hardware!

### DIFF
--- a/ur_description/urdf/ur10.transmission.xacro
+++ b/ur_description/urdf/ur10.transmission.xacro
@@ -6,7 +6,7 @@
     <transmission name="${prefix}shoulder_pan_trans">
       <type>transmission_interface/SimpleTransmission</type>
       <joint name="${prefix}shoulder_pan_joint">
-        <hardwareInterface>EffortJointInterface</hardwareInterface>
+        <hardwareInterface>PositionJointInterface</hardwareInterface>
       </joint>
       <actuator name="${prefix}shoulder_pan_motor">
         <mechanicalReduction>1</mechanicalReduction>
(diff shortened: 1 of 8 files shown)


gavanderhoorn:
> This would also solve #87

I would say it works around it, as having the actual parameters of a UR5/10 and being able to do dynamic simulations would be very nice.

Other than that, it might indeed make sense to do this, thanks for getting it done.

fmessmer:
True, it's only about getting rid of the PID "guessing".....rephrasing above, i.e. #87 should stay open...

abubeck:
This looks very good. I will merge this.

culletom:
@ipa-fxm Could I just ask whether similar changes can be made to the hydro-devel branch in order to use the PositionJointInterface, or is the above specific to indigo?

fmessmer:
according PR for `hydro-devel` can be found in #179

culletom:
@ipa-fxm Thank you, much appreciated!

PR #163, opened 2015-02-16
dynamic reconfigure server for prevent_programming

fmessmer:
Tested on UR5 v1.8
This works quite nice! Simply check the the parameter in `rqt_reconfigure` before you want to do anything with the robot (teach_mode, other movements from the teach pendant)....after you are done uncheck the parameter again and continue work on ROS side...


abubeck:
Looks good and was tested. Is only a minor change. I will merge this.

Issue #137, opened 2014-10-31
Procedure for bringing Hydro updates to Indigo


fmessmer:
There are currently several PullRequests open for Hydro (#128, #131, #135 and #136) as well as Indigo (#130 and #132).  

However, during the review of the Hydro PullRequests some more issues arose that leaded to updates and additional commits. (Some of those are not yet included in #130 and #132).

Still the Indigo PullRequest have been tested and are fine to be accepted.

In order to not get confused with all the various branches for the different features and ROS distributions respectively, I would like to propose the following procedure to bring the remaining (important!) updates from Hydro to Indigo:

After the PullRequests for Hydro (#128, #131, #135 and #136) and Indigo (#130 and #132) have been accepted and merged (as they are), I will compose a single new PR for bringing the remaining features from Hydro to Indigo. This PR will include (typo fixes in the README, material for the collision meshes,  the `max_velocity` parameter handling and the version-aware deserialization).

I hope you can agree to this procdure...

---

Furthermore, I would like to suggest that after this, we trigger new releases for both Hydro and Indigo and then switch the default branch of the repository to `indigo-devel`.

All upcoming new features (like the revision of the IO-Interface) will then only go into indigo-devel.
(Except for Hydro bug-fixes, of course!)

shaun-edwards:
Sounds good to me.

On Friday, October 31, 2014, Felix Messmer notifications@github.com wrote:

> There are currently several PullRequests open for Hydro (#128
> https://github.com/ros-industrial/universal_robot/pull/128, #131
> https://github.com/ros-industrial/universal_robot/pull/131, #135
> https://github.com/ros-industrial/universal_robot/pull/135 and #136
> https://github.com/ros-industrial/universal_robot/pull/136) as well as
> Indigo (#130 https://github.com/ros-industrial/universal_robot/pull/130
> and #132 https://github.com/ros-industrial/universal_robot/pull/132).
> 
> However, during the review of the Hydro PullRequests some more issues
> arose that leaded to updates and additional commits. (Some of those are not
> yet included in #130
> https://github.com/ros-industrial/universal_robot/pull/130 and #132
> https://github.com/ros-industrial/universal_robot/pull/132).
> 
> Still the Indigo PullRequest have been tested and are fine to be accepted.
> 
> In order to not get confused with all the various branches for the
> different features and ROS distributions respectively, I would like to
> propose the following procedure to bring the remaining (important!) updates
> from Hydro to Indigo:
> 
[...]

PR #164, opened 2015-02-16
driver: Specify explicit queue size when creating publishers.

de-vri-es:
This PR specifies the queue size when creating a publisher to silence warnings from ROS.

`universal_robot/ur_driver/src/ur_driver/driver.py:95: SyntaxWarning: The publisher should be created with an explicit keyword argument 'queue_size'. Please see http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers for more information.`

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -92,9 +92,9 @@ connected_robot_lock = threading.Lock()
 connected_robot_cond = threading.Condition(connected_robot_lock)
 last_joint_states = None
 last_joint_states_lock = threading.Lock()
-pub_joint_states = rospy.Publisher('joint_states', JointState)
-pub_wrench = rospy.Publisher('wrench', WrenchStamped)
-pub_io_states = rospy.Publisher('io_states', IOStates)
+pub_joint_states = rospy.Publisher('joint_states', JointState, queue_size=10)
+pub_wrench = rospy.Publisher('wrench', WrenchStamped, queue_size=10)
+pub_io_states = rospy.Publisher('io_states', IOStates, queue_size=10)
 #dump_state = open('dump_state', 'wb')
 
 class EOF(Exception): pass


tnl-woi:
Just a quick note:
I know this has been fixed in the Indigo branch, not in the Hydro branch. I wonder if there is a specific reason for _not_ doing this in the Hydro branch. (Maybe @ipa-fxm knows?)

Are you sure you have tested this in Hydro?

If the change is valid for Hydro, you might apply the same change in `test_comm.py` and `testRT_comm.py` (2 times in both files), and use the same values as in Indigo (`queue_size=1`, not `10`) for consistency.

de-vri-es:
Ah, thanks for the info. I wasn't aware this was fixed in the indigo branch. Indeed, I tested with indigo. I was under the impression that development was still done on the hydro branch and that the indigo branch was generally lagging behind. Is this still the case?

gavanderhoorn:
@tnl-woi: setting queue length on Hydro works as well. But it has probably only been fixed on `indigo-devel` as that is where the warnings are shown.

tnl-woi:
I thought main development is now done on indigo-devel, and only bug fixes go to hydro. See #137. Although, looking at the change logs, I think hydro and indigo are quite in sync. The queue_size argument is about the only change between these branches (which may be reason to apply this PR, by the way).

Given the answer of @gavanderhoorn, it seems the argument was introduced in Hydro and the warnings were introduced in Indigo. That explains why it was only changed in Indigo.

Strictly speaking there's no need to fix this in hydro, although it may be good for some people and doesn't do any harm, and minimize the differences between the hydro and indigo branches. I'm not sure what the policy is though.

de-vri-es:
> I thought main development is now done on indigo-devel, and only bug fixes go to hydro. See #137.

Thanks, looks like I should have been using the indigo branch. Oops. I just cloned and stayed on the branch I got.

> The queue_size argument is about the only change between these branches (which may be reason to apply this PR, by the way).

Then it's probably better to fully merge indigo and hydro branches rather than this PR, to keep things easy on git. If the explicit queue size really is the only difference that is. Seems like silly thing to keep two versions for.

Either way, this PR probably shouldn't be merged, so I'll close it.

PR #165, opened 2015-02-16
driver: Fix ParseError in UR script.

de-vri-es:
The last line in the driver UR script causes the robot controller to spit out an error message. Tested with the simulator versions 1.8, 3.0 and 3.1. I'm not sure about older versions, but I know it's not necessary to call your program after sending it. It is executed even if you don't. And apparently it's not just unnecessary but also not possible.

Does anyone know if this was required on older versions of the software? If not this PR fixes the parse errors given by the robot controller.

```
TCPServer: Secondary client connected
NameError: name 'driverProg' is not defined
--------------------------------------
TCPReceiver::parseLine: Parse Error   
driverProg()
--------------------------------------
```

### DIFF
--- a/ur_driver/prog
+++ b/ur_driver/prog
@@ -226,4 +226,3 @@ def driverProg():
   
   socket_send_int(MSG_QUIT)
 end
-driverProg()


de-vri-es:
I just noticed RESET_PROGRAM already doesn't include a call to `resetProg()` after the definition. AFAIK that's the correct way to do it. But if not, `resetProg()` should be modified to also include a call to itself after the definition.

gavanderhoorn:
Seeing as `prog` has included this since the beginning (of time, see 387bdaa1), I think it's safe to assume that it was indeed necessary on the older system versions.

de-vri-es:
I just tested with the oldest URSim I could get my hands on, which is 1.5.8077. That already executes the function if you don't call it, and it gives the same parse error if you do:

```
TCPServer: Secondary client connected
NameError: name 'main' is not defined
--------------------------------------
TCPReceiver::parseLine: Parse Error   
main()
--------------------------------------
TCPReceiver: Success
TCPSender: Broken pipe
TCPServer: Secondary client disconnected
```

My guess is that it was never necessary and actually incorrect (since the beginning of time)

de-vri-es:
I also had a look at the script manual for version 1.2 (the oldest manual I could find). While doesn't say anything about it explicitly, it seems to suggest that a user program doesn't need to be called (section 1.6 of the scriptmanual for version 1.2):

> A urscript program is declared as a function without parameters:
> 
> ```
> def myProg():
> end
> ```

It only says declared and not called, but all examples in the manual end without calling the user program explicitly. So, I really think the call to `driverProg()` should be dropped.

It feels natural that a function should be called to be executed, so I think it may have been added without much thought. But in the case of uploading URScript programs, it seems to be wrong.

gavanderhoorn:
Seems you're right (trusting something that has been like that 'since the beginning of time' is also not a very good argument for why it would be correct).

+1 to this change.

abubeck:
Thanks a lot for your effort. This message was annoying me already :)
I will merge this into hydro and also add it for indigo.

PR #166, opened 2015-02-16
driver: Factor out __send_message method.

de-vri-es:
This PR puts acquiring the required lock and packing the data in a separate method in CommanderTCPHandler. It has no functional changes. Quickly tested in simulator 3.0.

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -420,7 +420,7 @@ class CommanderTCPHandler(SocketServer.BaseRequestHandler):
                     raise EOF()
 
     def handle(self):
-        self.socket_lock = threading.Lock()
+        self.__socket_lock = threading.Lock()
         setConnectedRobot(self)
         print "Handling a request"
         try:
@@ -464,9 +464,15 @@ class CommanderTCPHandler(SocketServer.BaseRequestHandler):
             print "Connection closed (command):", ex
             setConnectedRobot(None)
 
+    # Send a message to the robot.
+    def __send_message(self, params):
+        # Acquire a lock first to prevent race conditions.
+        buf = struct.pack("!%ii" % len(params), *params)
+        with self.__socket_lock:
+            self.request.send(buf)
+
     def send_quit(self):
-        with self.socket_lock:
-            self.request.send(struct.pack("!i", MSG_QUIT))
+        self.__send_message([MSG_QUIT])
 
     def send_servoj(self, waypoint_id, q_actual, t):
         assert(len(q_actual) == 6)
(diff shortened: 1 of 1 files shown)


ur_driver/src/ur_driver/driver.py
```diff
@@ -464,9 +464,14 @@ def handle(self):
             print "Connection closed (command):", ex
             setConnectedRobot(None)
 
+    # Send a message to the robot.
+    def __send_message(self, params):
+        # Acquire a lock first to prevent race conditions.
+        with self.__socket_lock:
+            self.request.send(struct.pack("!%ii" % len(params), *params))
+
```
tnl-woi, line 473: I prefer to keep `buf = struct.pack(...)` in a separate line. There's no need to do this inside the lock, and I try to avoid putting multiple statements on one line. Just my two cents :)

tnl-woi:
I added one comment. Besides that, the change looks fine to me. :+1:

de-vri-es:
> I prefer to keep buf = struct.pack(...) in a separate line. There's no need to do this inside the lock, and I try to avoid putting multiple statements on one line. Just my two cents :)

Good point. Changed :)

gavanderhoorn:
@de-vri-es: could you perhaps add some comments to `__send_message(..)`? While this change does reduce code duplication, it also obscures the fact that all `params` are treated as integers in the call to `struct.pack(..)`.

de-vri-es:
I wanted to add a python docstring, but no other place in the code uses them. Shall I use regular comments instead then, for consistency?

de-vri-es:
Also, should I change this to be a pull request to indigo-devel?

gavanderhoorn:
No, just add docstrings. The rest of the code should be documented properly anyway, might as well start with this method.

re: branch: #137 says this:

> All upcoming new features (like the revision of the IO-Interface) will then only go into indigo-devel.
> (Except for Hydro bug-fixes, of course!)

While not a _new feature_, refactoring is also not a bug fix, so I guess this should go into `indigo-devel`. Perhaps @ipa-fxm has an opinion?

tnl-woi:
I agree with applying to the indigo-devel branch, but I'm not an authority on this point, just another user :-)

Regarding style, there is an official [style guide for ROS Python code](http://wiki.ros.org/PyStyleGuide), but indeed the driver code does not fully comply with that (at least the documenting style is different). In my opinion, new code should be added in the correct style, and old code should be converted as time permits. This will gradually improve the code style over time. Not sure if everybody agrees with this approach, though.

de-vri-es:
I don't believe I can change the pull request to apply to indigo-devel, so I'll close this one and open a new one.

PR #168, opened 2015-02-17
Included a description of how to use the arm_prefix parameter for the URXKinematics Plugin and Gazebo self collides

UltronDestroyer:


fmessmer:
I'm sorry, but I don't understand what this PR is supposed to solve?
You added launch file xml code that starts a node into a yaml file...

UltronDestroyer:
This adds some needed documentation on the arm_prefix parameter. For use inside the husky & grizzly platforms (and in order to use the custom URKinematics solver),  I had to pass in an arm_prefix. To find out that there was an arm_prefix parameter, I had to look through the source code. There was then nothing documenting how to set this arm prefix parameter. It turns out, inside a launch file from the config generated from moveit, I had to add that param to that namespace. I thought this method was a bit convoluted for what is actually being achieved here. 

The XML code is commented out. It would be great to see some foreward facing documentation on using the URX arms with custom robots. In the meantime, I've just added some in source comments on how to do that.

UltronDestroyer:
Please see the newest commit which adds self collides to the UR5 Gazebo model.

fmessmer:
First of all, you should not mix several features within the same PR...

On the other hand:
The `<selfCollide>` tags are not working as I had to find out myself (see my [post1](http://answers.gazebosim.org/question/7074/urdf-to-sdf-conversion-using-gzsdf/) and my [post2](http://answers.gazebosim.org/question/7082/using-sdf-tags-in-urdf-gazebo-extension-directly/) on answers.gazebo.org
The related gazebo/sdf_parser [issue](https://bitbucket.org/osrf/sdformat/issue/71) seem not to be fixed yet...

@TheDash, or are you able to make the meshes of the UR actually collide within gazebo (i.e. not one mesh interuding the other)?

UltronDestroyer:
It seems that I had trouble with MoveIt if these weren't enabled. (Issues with them colliding with the baselink)

fmessmer:
> The XML code is commented out.

No, it's not, as can be seen in https://github.com/TheDash/universal_robot/blob/hydro-devel/ur10_moveit_config/config/kinematics.yaml#L6

fmessmer:
> It seems that I had trouble with MoveIt if these weren't enabled. 

What kind of troubles?
MoveIt! should not even bring you into the situation where links collide - that's what MoveIt! is all about!

Please report an issue before describing the problem before are sending PullRequests...so that others might verify/help/solve the issue...

UltronDestroyer:
The base_link in my robot was colliding with the UR5 arm if the self collides weren't there

fmessmer:
Well are you using an additional base_link, i.e. do you have the UR mounted to some other link, i.e. box or mobile platform or what-so-ever?
If yes, you cannot use the ur5_moveit_config as it is only meant to be used with a UR5 solo!

UltronDestroyer:
Well yes, we are using an additional moveit_config. We just thought it would make more sense to include in our xacro a ur5.gazebo.xacro instead of making our own. Besides, the current ur5.gazebo.xacro isn't doing much anyway so let's make use of it :D

fmessmer:
It's not making much use of it...because the tags in there do not work (see the gazebo posts)....and we do not need materials as we are using collada files...

UltronDestroyer:
Hmm I will remove these tags from my code and see if I have any issues, and then get back to you.

UltronDestroyer:
This is the result without this code. http://imgur.com/Vpauawf
This is the result with the code.  http://imgur.com/9gm7765

UltronDestroyer:
ping @ipa-fxm 

Can we get this merged? I would like to prevent the UR5 arm from falling through robots. Let me know of any changes you'd like before this can happen.

tnl-woi:
I don't know about the Gazebo stuff, but:
- The XML code in `ur10_moveit_config/config/kinematics.yaml` is still _not_ commented out.
- `ur10_moveit_config/config/kinematics.yaml` contains some references to ur5 instead of ur10 (line 3 and 8).
- The XML code in `ur5_moveit_config/config/kinematics.yaml` contains invalid hashes at line 3, 6 and 8.
- I think it's a good idea to add some documentation but I'd prefer that in a text file somewhere, instead of in a configuration file. (This also has the advantage that you don't have to manage the documentation at two places.)
- Why is only `ur5.gazebo.xacro` modified, and not `ur10.gazebo.xacro`?

UltronDestroyer:
- Why is only ur5.gazebo.xacro modified, and not ur10.gazebo.xacro?
  *\* See 8b05bf5
- I think it's a good idea to add some documentation but I'd prefer that in a text file somewhere, instead of in a configuration file. (This also has the advantage that you don't have to manage the documentation at two places.)
  *\* See e4c4508
- The XML code in ur5_moveit_config/config/kinematics.yaml contains invalid hashes at line 3, 6 and 8.
  *\* See e4c4508
- ur10_moveit_config/config/kinematics.yaml contains some references to ur5 instead of ur10 (line 3 and 8).
  *\* See e4c4508
- The XML code in ur10_moveit_config/config/kinematics.yaml is still not commented out.
  *\* See e4c4508

That should be all that is needed for now.. may add more commits later if bugs are found.

tnl-woi:
Thanks for the updates, Devon. As I said before, I'm not familiar with Gazebo, so I can't "vote" for this PR. But at least I can help reviewing some of the changes to remove issues that are probably in the way for a merge.

Documentation (`README.MD`) seems fine to me (a minor style issue though: the `roslaunch` command should be marked as code).
`ur10_moveit_config/config/kinematics.yaml` is still broken and `ur5_moveit_config/config/kinematics.yaml` only contains whitespace changes. Can you just revert those two files to what they were?

The other files look good to me. I'll leave it to others that are more familiar with Gazebo to accept the PR (or not).

A related question, if I may: is it possible to control multiple UR5/UR10 robots with this driver? I guess the prefixes are mainly intended to distinguish between a "left" and "right" arm, for example. Does the driver support this already? The reason that I ask this is that at least the IO interface does not seem to support this (it uses global variables in the ur_driver.io_interface module). If it doesn't work indeed, I think we should create a new issue for this. Should I create a new issue for this, or does it already work?

gavanderhoorn:
@tnl-woi wrote:

> A related question, if I may: is it possible to control multiple UR5/UR10 robots with this driver? I guess the prefixes are mainly intended to distinguish between a "left" and "right" arm, for example. Does the driver support this already? The reason that I ask this is that at least the IO interface does not seem to support this (it uses global variables in the ur_driver.io_interface module). If it doesn't work indeed, I think we should create a new issue for this. Should I create a new issue for this, or does it already work?

Depends on what you mean with 'multiple robots' exactly.

If you are referring to multiple motion groups (something the UR ctrlr does not support afaik), then no, the current implementation assumes there is only one (which is true). For the case where you have multiple controllers each controlling their own arm, then the idea is to just instantiate the driver node(s) multiple times, combined with suitable TF prefixes and node namespaces.

The driver itself does not need any specific support for multiple controllers.

README.md
```diff
@@ -44,6 +44,13 @@ For starting up RViz with a configuration including the MoveIt! Motion Planning
 
 ```roslaunch ur5_moveit_config moveit_rviz.launch config:=true```
 
+To use in conjunction or as an addition with another robot, you'll need to modify the arm_prefix parameter that will be used in RViz.
```
gavanderhoorn, line 47: isn't it _in conjunction with_ and _an addition to_?

ur5_moveit_config/launch/moveit_rviz.launch
```diff
@@ -11,6 +12,7 @@
   <node name="$(anon rviz)" launch-prefix="$(arg launch_prefix)" pkg="rviz" type="rviz" respawn="false"
 	args="$(arg command_args)" output="screen">
     <rosparam command="load" file="$(find ur5_moveit_config)/config/kinematics.yaml"/>
+    <param name="arm_prefix" value="$(arg arm_prefix)"/>
```
gavanderhoorn, line 15: Is this an actual RViz parameter? The Robot Model display has a _TF Prefix_ field, but I've never been able to set that from a ros parameter.

A quick test shows RViz happily ignoring it, on my local machine.

UltronDestroyer, line 15: This is required because the URKinematics plugin is dynamically linked to the RViz process. The URKinematics plugin looks for the private parameter arm_prefix, which has to be set inside the RViz namespace.

gavanderhoorn, line 15: Could you be more specific about what you mean with "is dynamically linked to the RViz process"? Do you mean `move_group` node perhaps? If the latter is the case, shouldn't starting the `move_group` node in a suitable namespace be enough to get it to work with a particular prefix?

UltronDestroyer, line 15: The kinematics.yaml tells RViz what plugin to use. Since RViz is spawned in this stanza of the launch file, it is the "parent". The URKinematics plugin is thus equal to the namespace of RViz, and will look for ~params (Private params) with respect to its namespace (in this case it is RVIz because it is a plugin).

So the param, if put anywhere else but the RViz namespace, won't be readable by the URKinematics plugin (because of how it is written in code).

See (Line 326)

```
 private_handle.param<std::string>("arm_prefix", arm_prefix_, "");
```

in https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur_kinematics/src/ur_moveit_plugin.cpp

gavanderhoorn, line 15: > The URKinematics plugin is loaded into RViz when the kinematics.yaml is parsed

I'm confused: afaik, kinematics plugins are loaded by MoveIt (from [moveit/Kinematics/C++ API](http://moveit.ros.org/wiki/Kinematics/C%2B%2B_API)):

> The `RobotModel` and `RobotState` classes are the core classes that give you access to the kinematics. [..]

The `kinematics.yaml` file is loaded onto the parameter server by roslaunch, and the parameters are then [used by MoveIt](http://moveit.ros.org/wiki/Kinematics/Configuration) to configure the referenced subclasses of `KinematicsBase`.

MoveIt also exposes the kinematics solvers in one of its [capabilities](http://moveit.ros.org/wiki/Capabilities#Kinematics).

RViz is not involved in this, other than that it provides an interface to most of the MoveGroup functionality (through the assorted services / action servers).

I don't currently know where that private `NodeHandle` in `URKinematicsPlugin` resolves to, but I don't really understand how it could be RViz?

UltronDestroyer, line 15: After running this code,

dash@Dash:~/development_ws$ rosparam list | grep prefix
/rviz_Dash_5364_7703676098926058836/arm_prefix

which obviously, by the code shows it's in RViz namespace,

since the URKinematics looks for the arm_prefix parameter, you have to wonder where else could it be getting set from? 

When I remove that LOC from moveit_rviz.launch:

[ERROR] [1424975775.985587915, 21.224000000]: Kin chain provided in model doesn't contain standard UR joint 'shoulder_lift_joint'.
[ERROR] [1424975775.985692288, 21.224000000]: Kinematics solver of type 'ur_kinematics/UR5KinematicsPlugin' could not be initialized for group 'ur5_arm'
[ERROR] [1424975775.986159403, 21.224000000]: Kinematics solver could not be instantiated for joint group ur5_arm.

Even when adding it to another place in moveit_rviz.launch, or adding it to kinematics.yaml, I still received that error. It has to be in the rviz namespace or else the kinematics plugin doesn't work, since the kinematics plugin looks for the arm prefix in its current ns, it must mean its in the RViz NS

The only way to fix it is to add that parameter at that specific location. I'm not sure exactly how moveit_plugins are loaded or where they're loaded into.. I've taken a look but haven't seen anything showing where moveit plugins are loaded in yet.

fmessmer:
Sorry guys, as I'm not in the office at the moment, I'm not able to follow all the discussions going on here....I would like to test some things and then I will get to you...

ur5_moveit_config/launch/moveit_rviz.launch
gavanderhoorn, line 15: > The only way to fix it is to add that parameter at that specific location. I'm not sure exactly how moveit_plugins are loaded or where they're loaded into.. I've taken a look but haven't seen anything showing where moveit plugins are loaded in yet.

The plugins can be loaded twice (in a normal MoveIt config pkg): once by `move_group`, and once by (I guess) the RViz MoveIt plugin. As `URKinematicsPlugin::initialize(..)` instantiates a private `NodeHandle`, this makes the instance initialised by RViz look for parameters under the `/rviz_whatever_../` ns. The other instance looks for the parameter under `/move_group/`.

I'm not entirely sure how things are supposed to work actually, especially with `kinematics.yaml` being loaded under the `$(arg robot_description)_kinematics` namespace (by `planning_context.launch`), which is completely ignored by IK plugins, afaict.

However, I think adding the `arm_prefix` parameter to `kinematics.yaml` (at root level, or perhaps even a per-group property) would be much cleaner (also: how is the plugin otherwise supposed to retrieve this parameter if only `move_group.launch` is started (as is the case on _headless_ robots)).

ur_driver/test_move.py
```diff
@@ -91,6 +93,10 @@ def main():
     global client
     try:
         rospy.init_node("test_move", anonymous=True, disable_signals=True)
+        prefix = rospy.get_param("/ur_driver/prefix", "")
```
gavanderhoorn, line 96: This should probably just read the parameter in the nodes private namespace, instead of hardcoding it like this.

fmessmer:
To be honest:
This PR is a mess...several features...a lot of subsequent commits that are not squashed...still remarks from the community not considered....Thus, I vote for closing this!!!

To deal with the features separately, I started to create a new PR for the gazebo tag `selfCollide` (see #186)
@TheDash please test this branch and comment in #186 whether this works for you!

As to the second "feature" of this PR, i.e. `arm_prefix` handling:
I'm still not clear why you want to use the urX_moveit_config provided in this repository for a robot that consists of a mobile base and the UR as a component...moveit_configs are setup-specific...and using the UR on top of a mobile base definitely differs from using the UR stand-alone.
For example the self-collision matrix in the *.srdf does not reflect your setup at all.
Thus, it is common practise to create a new moveit_config for each setup respectively!
As this is very easy using the MoveIt!-SetupAssistant, this should not be too much trouble...
And: as a nice side-effect, the newly created moveit_config for your setup will have kinematics.yaml files with the correct joint_names for the arm!!!

However, being able to use the `test_move.py` considering an `arm_prefix` parameter seems reasonble to me! @TheDash could you create a new PR with just the changes you did in the `test_move.py` script?

UltronDestroyer:
Yeah, I'm going to close this and re-open a cleaner PR.

ur5_moveit_config/launch/moveit_rviz.launch
UltronDestroyer, line 15: Coming back to this a year later:

We're now making a Husky with 2 UR5 arms on it and I'm running into a similar issue. The issue this time is, how to load 2 rms with different prefixes. is it possible to load separate move groups for RViz? I'm guesssing by namespacing the move_group plugin I should be able to achieve this. @gavanderhoorn

gavanderhoorn, line 15: I asked about this on `moveit-users` some time ago, see [parameters for ikfast plugins using private NodeHandles](https://groups.google.com/d/msg/moveit-users/jlneWn9uumo/aceglJNfkUwJ). The response by @sachinchitta is [here](https://groups.google.com/d/msg/moveit-users/gk1lPiP8ylo/85Ba9BVKptkJ).

gavanderhoorn, line 15: But with multiple instances of the same plugin it might not work as described/expected. See also discussion in #248.

Issue #162, opened 2015-02-13
'prevent_programming' should be set using dynamic_reconfigure, not by polling the parameter server


gavanderhoorn:
Even though it's only [accessed](https://github.com/ros-industrial/universal_robot/blob/1d99ea831dae0259634d9389c1ee42690e9c3d82/ur_driver/src/ur_driver/driver.py#L968-L974) when the Python side loses connection to the controller (and on initial upload), polling the parameter server for the value of `prevent_programming` doesn't seem very nice. 

The proper alternative would seem to be adding a dynamic_reconfigure server to the driver.

fmessmer:
I started this on [my branch](https://github.com/ipa-fxm/universal_robot/tree/prevent_programming).
I need to test it with the hardware - thus, no PR yet...

fmessmer:
see #163

fmessmer:
see #170

abubeck:
Dynamic reconfigure is in (#163). Therefore I will close this.
Discussion on having both (rosparam and dynamic_reconfigure) is in #170

PR #170, opened 2015-02-20
[Indigo] enable reconfigure and set_param side by side

fmessmer:
In PR #163 it was not possible to set the parameter `prevent_programming` via `rosparam set /ur_driver/prevent_programming true`.
Now both ways (rosparam and dynamic_reconfigure) can be used side by side

Sorry about that...

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -894,11 +894,7 @@ def main():
         rospy.logwarn("use_sim_time is set!!!")
     
     global prevent_programming
-    prevent_programming = rospy.get_param("prevent_programming", False)
     reconfigure_srv = Server(URDriverConfig, reconfigure_callback)
-    ## Still use parameter server?
-    #update = URDriverConfig(prevent_programming)
-    #reconfigure_srv.update_configuration(update)
     
     prefix = rospy.get_param("~prefix", "")
     print "Setting prefix to %s" % prefix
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Doesn't this reintroduce the polling of the parameter server? Why would we want that?

fmessmer:
Well, when using dynamic reconfigure for `prevent_programming`, the parameter still appears on the ParameterServer (`rosparam list`).
Without the changes proposed here, someone who tries to change the parameter using `rosparam set` will be surprised that it has no effect...
With the changes proposed here, both options (rqt_reconfigure and `rosparam set`) can be used!

gavanderhoorn:
> Without the changes proposed here, someone who tries to change the parameter using `rosparam set` will be surprised that it has no effect...

I'm confused: aren't `NodeHandle::getParam(..)` & friends meant for accessing parameters mostly at node initialisation, and Dynamic Reconfigure for dealing with changes to parameters _at runtime_?

As far as I have always understood, `rosparam set ..` was never meant to affect nodes after they have started up. You could perhaps use the [caching versions](http://wiki.ros.org/roscpp/Overview/Parameter%20Server#Cached_Parameters) of `getParam(..)`, but why bother with dynamic reconfigure in that case in the first place.

fmessmer:
> I'm confused: aren't NodeHandle::getParam(..) & friends meant for accessing parameters mostly at node initialisation

Well, maybe that's what they are meant for...but before my [commit](https://github.com/ros-industrial/universal_robot/commit/62b9c5c74a56e17bf437f0efba003cab843479e0) has been merged, there was a continuous call to `prevent_programming = rospy.get_param("prevent_programming", False)` during the main loop `while not rospy.is_shutdown():`
Thus, `rospy.get_param` has been used as a quasi-dynamic_reconfigure....although it was querying the parameter `/prevent_programming` instead of `/ur_driver/prevent_programming`

...and now, you can simply do both!

gavanderhoorn:
> Thus, `rospy.get_param` has been used as a quasi-dynamic_reconfigure

Exactly. That was wrong, and was what #162 was all about.

abubeck:
The dynamic reconfigure part has been done, merged and tested. (#163)
This only adds the "old api" based on rosparam. But the parameter is now under the namespace of the driver. As we want to be backward compatible I think it makes sense to have both. Therefore I would prefer to also merge this one.

abubeck:
Will merge this now.

PR #172, opened 2015-02-23
driver: Factor out __send_message method.

de-vri-es:
Same as #166 but based on indigo-devel and with python docstring added. It may be a good idea do determine what style of docstring to use for future documentation. There are a number of conventions, but not one clear winner AFAIK. Since this docstring would set a precedent for the driver it may be a good idea to nitpick about it.

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -423,7 +423,7 @@ class CommanderTCPHandler(SocketServer.BaseRequestHandler):
                     raise EOF()
 
     def handle(self):
-        self.socket_lock = threading.Lock()
+        self.__socket_lock = threading.Lock()
         setConnectedRobot(self)
         print "Handling a request"
         try:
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
I suggest we keep the same style as that used in other ROS Python source. Some examples: [genmsg/gentools.py](https://github.com/ros/genmsg/blob/indigo-devel/src/genmsg/gentools.py) and [python/catkin/package_version.py](https://github.com/ros/catkin/blob/indigo-devel/python/catkin/package_version.py).

ur_driver/src/ur_driver/driver.py
```diff
@@ -467,9 +467,23 @@ def handle(self):
             print "Connection closed (command):", ex
             setConnectedRobot(None)
 
+    def __send_message(self, data):
+        """
+        Send a message to the robot.
+
+        The message is given as a list of integers that will be packed
+        as 4 bytes in network byte order (big endian).
```
gavanderhoorn: This might be just me, but _that_ seems to be referring to _a list_, which then would mean that the entire list is packed into 4 integers?

de-vri-es: Heh, indeed. Will change that.

de-vri-es:
Docstring updated.

abubeck:
Nice cleanup. I have tested this and it works well. Therefore I will merge this in.

PR #174, opened 2015-02-25
Prefix addition

UltronDestroyer:
Adds a prefix to the test_move.py script. Otherwise, a robot that has it defined with a prefix will be unable to run the test_move script without copying the code and modifying it to their own directory. This keeps the code modular.


UltronDestroyer:
Oops, stacked multiple PRs.

PR #175, opened 2015-02-25
Added prefix parameter to test move script.

UltronDestroyer:


UltronDestroyer:
Oops, didn't make it so it added only one commit from the main branch.

Issue #155, opened 2015-01-23
Question about IKFast & UR arms


UltronDestroyer:
Does there exist a working implementation of OpenRave's IKFast system for UR10 & UR5 arms? Just curious before I start hacking away on one!

shaun-edwards:
We don't have one for a bare arm (all of ours have an end-effector on them).  I can say that ikfast is an order of magnitude faster than KDL.

UltronDestroyer:
I see. And yes, KDL is causing some difficulties within MoveIt. Do you know if there's an available IKFast for MoveIt? I saw the moveit_ikfast package, but is there any kind of configuration I'll need to do for the UR10?

shaun-edwards:
It's probably easiest to re-run the moveit setup assistant on the existing package.  The drop down for kinematic chains should show any/all IK plugins.

Otherwise, you could edit the `kinematics.yaml` file to point to the fast IK plugin package.  The fanuc packages make use of ikfast (see [here](https://github.com/ros-industrial/fanuc/blob/hydro-devel/fanuc_lrmate200ic5l_moveit_config/config/kinematics.yaml)).

UltronDestroyer:
Ah ok, so I do have to configure/create one for the UR10 arm then.

gavanderhoorn:
@TheDash: does it have to be ikfast based? The repository already contains IK/FK packages for both the UR5 and the UR10. In principle I'd say they should offer the same (or even better, see [here](http://openrave-users-list.185357.n3.nabble.com/OpenRAVE-IKFast-Failure-UR10-td4026158.html)) level of functionality, if we can get the issues that some users have reported cleared up (#127).

It would perhaps be more efficient to try and fix that, than starting from scratch with IKFast (which can be difficult for the URs).

fmessmer:
In the [kinematics.yaml](https://github.com/ros-industrial/universal_robot/blob/hydro-devel/ur5_moveit_config/config/kinematics.yaml) there is already a commented block that would use the analytical IK solver from the [ur_kinematics](https://github.com/ros-industrial/universal_robot/tree/hydro-devel/ur_kinematics) package (it's also there for the UR10).

You can try it, but as @gavanderhoorn mentioned this solver will return IK solutions that are within the full joint range [-2Pi...2Pi] so that it only works with the ur_description with the limited option set to false, with which MoveIt! cannot plan in a reasonalbe amount of planning time....

[...]

PR #184, opened 2015-03-06
[Indigo] crop ik solutions wrt joint_limits

fmessmer:
This PR crops the ik_solutions provided by ur_kinematics wrt the joint_limits used in the `/robot_description`.
Thus, the same plugin now works for the limited version of the UR5/UR10 as well!

### DIFF
--- a/ur10_moveit_config/config/kinematics.yaml
+++ b/ur10_moveit_config/config/kinematics.yaml
@@ -1,4 +1,3 @@
-## the UR10KinematicsPlugin only works with the unlimited UR10, i.e. limited:=false
 #manipulator:
 #  kinematics_solver: ur_kinematics/UR10KinematicsPlugin
 #  kinematics_solver_search_resolution: 0.005
--- a/ur5_moveit_config/config/kinematics.yaml
+++ b/ur5_moveit_config/config/kinematics.yaml
@@ -1,4 +1,3 @@
-## the UR5KinematicsPlugin only works with the unlimited UR5, i.e. limited:=false
 #manipulator:
 #  kinematics_solver: ur_kinematics/UR5KinematicsPlugin
 #  kinematics_solver_search_resolution: 0.005
--- a/ur_kinematics/src/ur_moveit_plugin.cpp
+++ b/ur_kinematics/src/ur_moveit_plugin.cpp
@@ -645,14 +645,57 @@ bool URKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose &ik_pose,
     // Do the analytic IK
     num_sols = inverse((double*) homo_ik_pose, (double*) q_ik_sols, 
                        jnt_pos_test(ur_joint_inds_start_+5));
+    
+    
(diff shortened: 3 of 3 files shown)


tnl-woi:
Tested with limited joints and works fine! :+1:
As for me, the URKinematics plugin may become the default again.

UltronDestroyer:
Does this work for unlimited?

fmessmer:
Yes, the same plugin now works for both limited and unlimited versions!

abubeck:
As this is tested. I will merge.

gavanderhoorn:
@abubeck: I think @TheDash has not yet reported either success or failure (and commented on #155). Seems premature to merge this without confirmation that it works for the user who reported the issue?

abubeck:
@gavanderhoorn : @ipa-fxm has tested it with both limited and unlimited versions.

gavanderhoorn:
Yes, I know :). But the `URKinematics` plugin was tested in the past as well, but then it didn't work on other peoples' setups. So some additional testing wouldn't seem to be a bad thing.

UltronDestroyer:
I just made an attempt to run the planner/solver with joint_limited:=true and false. The limited:=true makes the arm purple and only plannable by random configurations. The limited:=false never solves the freeform orb problem.

UltronDestroyer:
Note: this has been solved.. will report later with the solution.

Edit:

The issue was a combination of the changes made regarding the hardware interface and the delay of a package release.

I had used the deb package of UR which didn't have the kinematic planner changes to allow free planning motion. That didn't work, and I tried installing the source package instead. the source package however wasn't compatible with the controller interface for Gazebo that I had specified, and it wasn't reporting an error. Once I downloaded and used the universal_robot package from source, and changed my gazebo controller interface to be position_controllers, everything worked fine.

fmessmer:
@TheDash 
True, this feature is not yet released....also see my comments in #202

UltronDestroyer:
Right, it should be released because using things from source on all of the platforms/tutorials that I have put out can be problematic for users. For example, when I tried to install code on one of our platforms, to my dismay, I had to find out that you guys changed the hardware interface which affected the simulator and took me about 8 hours to fix. The only way I found that this was the actual change was by reading through the code commits on your personal fork of the project. 

Not only that, but the UR5KinematicsPlugin requires the joint configuration to be unlimited, and the MoveIt configurations requires the URDF to have it specified as the limited configuration so it is really confusing for myself and users.

When you re-release this package, can we somehow notify everyone that uses the UR5 arm in gazebo that their simulators will no longer work and tell them what the fix is? (E.g telling their configs to use position_controllers/PositionJointInterface). I only found that out by reading a commit message, which is shrouded away in the forests of github.

achim-k:
With this PR being merged, is there still a reason not to set the default IK-plugin back to `ur_kinematics/URXKinematicsPlugin`. The default was set to KDLPlugin in https://github.com/ros-industrial/universal_robot/pull/128
Sorry to bring this up again

PR #185, opened 2015-03-10
[Indigo] allow to start gazebo without gui

fmessmer:
accept additional launch file argument `gui` allowing to start gazebo simulation without GUI
(reduces cpu load on older machines :wink:)

### DIFF
--- a/ur_gazebo/launch/ur10.launch
+++ b/ur_gazebo/launch/ur10.launch
@@ -2,11 +2,13 @@
 <launch>
   <arg name="limited" default="false"/>
   <arg name="paused" default="false"/>
-
+  <arg name="gui" default="true"/>
+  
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Would it perhaps be an idea to name the parameter `use_gui`, like the `joint_state_publisher` one?

Although, if `gui` is standard for Gazebo, then that is also fine.

fmessmer:
The parameter you pass to gazebo is called `gui` (see [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/hydro-devel/gazebo_ros/launch/empty_world.launch#L6))

abubeck:
Looks good to me and doesn't affect other things. I will merge.

Issue #183, opened 2015-03-05
Fatal error when unpacking RobotState packet


akgoins:
When launching ur5_bringup.launch, I keep getting the error "Exception: Fatal error when unpacking RobotState packet" which prevents me from receiving any information on the "/io_states" topic.  I am getting data on the /joint_states and /wrench topics, and I can set IO using the /set_IO service, but I am not receiving any feedback.  Is this an issue with newer controllers?  I have interface and controller version 3.1.16916.  Here is the full debug output:

```
[WARN] [WallTime: 1425584960.972983] No calibration offset for joint "shoulder_pan_joint"
[WARN] [WallTime: 1425584960.973292] No calibration offset for joint "shoulder_lift_joint"
[WARN] [WallTime: 1425584960.973510] No calibration offset for joint "elbow_joint"
[WARN] [WallTime: 1425584960.973790] No calibration offset for joint "wrist_1_joint"
[WARN] [WallTime: 1425584960.974083] No calibration offset for joint "wrist_2_joint"
[WARN] [WallTime: 1425584960.974338] No calibration offset for joint "wrist_3_joint"
[INFO] [WallTime: 1425584960.974532] No calibration offsets loaded from urdf
[INFO] [WallTime: 1425584960.975878] Max velocity accepted by ur_driver: 10.0 [rad/s]
[INFO] [WallTime: 1425584960.979008] Bounds for Payload: [0.0, 10.0]
Disconnected.  Reconnecting
[INFO] [WallTime: 1425584961.013669] Programming the robot
[INFO] [WallTime: 1425584961.013935] Programming the robot at 192.168.1.50
Likely a syntax error:
4���������  URControlBFeb 06 2015, 15:17:07
Exception in thread URConnection:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/swri/catkin_ws/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 279, in __run
    self.__on_packet(packet)
  File "/home/swri/catkin_ws/src/universal_robot/ur_driver/src/ur_driver/driver.py", line 191, in __on_packet
    state = RobotState.unpack(buf)
  File "/home/swri/catkin_ws/src/universal_robot/ur_driver/src/ur_driver/deserialize.py", line 326, in unpack
[...]

PR #187, opened 2015-03-10
[Indigo] exception handling when unpacking RobotState

fmessmer:
This is related to #183 

@akgoins please test!

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -188,7 +188,12 @@ class URConnection(object):
         log("Halted")
 
     def __on_packet(self, buf):
-        state = RobotState.unpack(buf)
+        try:
+            state = RobotState.unpack(buf)
+        except Exception as e:
+            rospy.logerr("Exception while unpacking RobotState")
+            return
+        
         self.last_state = state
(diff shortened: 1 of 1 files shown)


ur_driver/src/ur_driver/driver.py
```diff
@@ -188,7 +188,12 @@ def __trigger_halted(self):
         log("Halted")
 
     def __on_packet(self, buf):
-        state = RobotState.unpack(buf)
+        try:
+            state = RobotState.unpack(buf)
+        except Exception as e:
+            rospy.logerr("Exception while unpacking RobotState")
```
gavanderhoorn, line 194: Technically this is incorrect (at least in the case of #183): the pkt received there is not a `ROBOT_STATE`, but a `ROBOT_MESSAGE`. Properly dealing with that would require more extensive changes though.

fmessmer:
I guess I thought the ROBOT_MESSAGE packet would only be received once...but it's always received as first part of the message, right?

I'll close this and try to find a correct solution asap...

gavanderhoorn:
Yes, if I understand the wiki correctly, `ROBOT_MESSAGE` is send only once, upon initial connection to the secondary interface (30002) (not sure about 30003).

PR #189, opened 2015-03-12
port of PR #157 to indigo

abubeck:
I have tested this on our UR5 with v1.8.
As it works for hydro I will merge this directly if nobody opposes.


fmessmer:
looks obviously the same as in hydro :wink:

abubeck:
As nobody opposes. I will merge now.

Issue #95, opened 2014-08-27
Add 'base' coordinate system link to URDFs


gavanderhoorn:
As discussed in ros-industrial/ros_industrial_issues#24.

gavanderhoorn:
I've recently done this myself, and I've got values that work. For `base`, only a rotation over Z wrt `base_link` is required, as the meshes in `ur_description` seem to have their origins rotated when compared with the real robot.

Should the meshes perhaps be updated? Then the `base` coordinate frame would be coincidental with `base_link` and things would be much more transparent.

Same goes for `tool0`.

fmessmer:
@abubeck has done this, too....umpf....
Could both of you create PullRequests?

gavanderhoorn:
@ipa-fxm wrote:

> @abubeck has done this, too [..]

I know, I spoke with him, but his branch uses even more rotations. There is only one needed, but even that one can be made unnecessary if we 'fix' the meshes (and thus the orientation of `base_link`).

gavanderhoorn:
#49 (_Add tool0 to URDFs_) is also related.

gavanderhoorn:
I think it would make sense to just rotate the `base_link` link `pi` rad over Z, then rotate the `visual` and `collision` geometry `-pi` to compensate. The `base` frame can then be coincident with `base_link`, no rotations needed anymore.

We can rotate the meshes later then.

fmessmer:
Well, one thing that needs to be considered about your proposed solution is the following:
UR robots are also used in composition with e.g. mobile platforms (e.g. Fraunhofer IPA's rob-@-work) or within environments. They also might exist in composition with tools/grippers attached in ROS. Thus, rotating the base_link in the ur_description will change all those setups!

What we did in @abubeck's solution is add new links `base0` and `tool0` like this:
`base_link` ->TransBase1 -> `base0` ->TransBase2 -> other_ur_links -> TransTool1 -> `tool0` -> TransTool2 -> `arm_ee_link`, where TransXXX are correctly set origins within additional fixed joints....

You can find it [here](https://github.com/abubeck/universal_robot/blob/feature/cartesian_motion/ur_description/urdf/ur5.urdf.xacro)...but I think that file was still rather WIP than final....

[...]

Issue #49, opened 2014-03-28
Add tool0 to URDFs


shaun-edwards:
The `tool0` frame should be added to the UR10 & 5 URDFs.  See ros-industrial/ros_industrial_issues#24.

gavanderhoorn:
Duplicate of #34.

fmessmer:
The according discussion now takes place in #95

Issue #80, opened 2014-07-07
driver: published WrenchStamped msg does not set frame name


gavanderhoorn:
The recently merged `WrenchStamped` publisher [does not](https://github.com/ros-industrial/universal_robot/blob/f003b5cd0f4ff9e8767e3b2d2fe4150f84b6943a/ur_driver/src/ur_driver/driver.py#L308-L322) seem to initialise the _header.frame_ field to any value. The scripting manual (v1.8, aug 2013) of the UR says:

> **get_tcp_force()**
> [..]
> The function returns ”p[Fx (N), Fy(N), Fz(N), TRx (Nm), TRy (Nm), TRz (Nm)]”. where Fx, Fy, and Fx are the forces in the axes of the robot base coordinate system measured in Newtons, and TRx, TRy, and TRz are the torques around these axes measyred in Newton times Meters.

(sic).

If done on purpose (not setting the frame name), then this should probably be documented. Otherwise, the correct frame should be set by the publisher.

gavanderhoorn:
Ping

abubeck:
We will work on this during merging of PR #78, but we want to wait until #92 is merged (tests pending).

fmessmer:
@Equanox @gavanderhoorn 
Is it correct that the `robot base coordinate system` is `${prefix}base_link` in our case?

gavanderhoorn:
yes, well, sort of. See ros-industrial/ros_industrial_issues#24 for some discussion.

The UR urdfs don't have the additional `base` link yet (#95), which means that `base_link` does not necessarily have the correct orientation wrt the real robot. Setting `base_link` as the `header.frame` link would then not work (I think).

@abubeck and @ipa-fmw should recognise this. See also [this](https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/WHbKhHjn_AM) discussion on the ROS-Industrial mailing list about it.

I have seen some discussion with UR about this, but I don't know whether we have been able to resolve this yet.

gavanderhoorn:
@ipa-fxm wrote:

> Is it correct that the `robot base coordinate system` is `${prefix}base_link` in our case?

As #200 has been merged, we should now be able to consider `${prefix}_base` as the `robot base coordinate system`.

gavanderhoorn:
`ur_driver` is deprecated and will not see further development.

Closing.

PR #200, opened 2015-04-22
Fix for issues #49 and #95: ros-i compatible base and tool0 frames

gavanderhoorn:
As per subject.

Quote of commit msg of db6b163:

> Note that `base` is essentially `base_link` but rotated by 180 degrees over the Z-axis. This is necessary as the visual and collision geometries appear to also have their origins rotated 180 degrees wrt the real robot.
> 
> `tool0` is similar to `ee_link`, but with its orientation such that it coincides with an all-zeros TCP setting on the UR controller. Users are expected to attach their own TCP frames to this frame, instead of changing it to match the physical setup.

Note also that the two frames are siblings of `base_link` and `ee_link`, not parents or children, so the kinematic chain `base_link->ee_link` is untouched. Existing MoveIt configurations and IKFast plugins should not be affected by these changes.

User-created composite urdfs (which instantiate either the `ur5_robot` or `ur10_robot` macros) could be affected, but only if they already have a link named `base` or `tool0` in their scene and they did not use a `prefix` when calling the robot macros (that could be considered user error).

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -300,5 +300,25 @@
     <xacro:ur10_arm_transmission prefix="${prefix}" />
     <xacro:ur10_arm_gazebo prefix="${prefix}" />
 
+    <!-- ROS base_link to UR 'Base' Coordinates transform -->
+    <link name="${prefix}base"/>
+    <joint name="${prefix}base_link-base" type="fixed">
+      <!-- NOTE: this rotation is only needed as long as base_link itself is
+                 not corrected wrt the real robot (ie: rotated over 180
+                 degrees)
+      -->
+      <origin xyz="0 0 0" rpy="0 0 ${-pi}"/>
+      <parent link="${prefix}base_link"/>
+      <child link="${prefix}base"/>
+    </joint>
+
+    <!-- Frame coincident with all-zeros TCP on UR controller -->
+    <link name="${prefix}tool0"/>
+    <joint name="${prefix}wrist_3_link-tool0" type="fixed">
+      <origin xyz="0 ${wrist_3_length} 0" rpy="${pi/-2.0} 0 0"/>
+      <parent link="${prefix}wrist_3_link"/>
+      <child link="${prefix}tool0"/>
+    </joint>
+
   </xacro:macro>
 </robot>
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -302,7 +302,7 @@
 
     <!-- ROS base_link to UR 'Base' Coordinates transform -->
     <link name="${prefix}base"/>
-    <joint name="${prefix}base_link-base" type="fixed">
+    <joint name="${prefix}base_link-base_fixed_joint" type="fixed">
       <!-- NOTE: this rotation is only needed as long as base_link itself is
                  not corrected wrt the real robot (ie: rotated over 180
                  degrees)
@@ -314,7 +314,7 @@
 
     <!-- Frame coincident with all-zeros TCP on UR controller -->
     <link name="${prefix}tool0"/>
-    <joint name="${prefix}wrist_3_link-tool0" type="fixed">
+    <joint name="${prefix}wrist_3_link-tool0_fixed_joint" type="fixed">
       <origin xyz="0 ${wrist_3_length} 0" rpy="${pi/-2.0} 0 0"/>
       <parent link="${prefix}wrist_3_link"/>
       <child link="${prefix}tool0"/>
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -42,7 +42,7 @@
 
   <!-- Arbitrary offsets for shoulder/elbow joints -->
   <property name="shoulder_offset" value="0.220941" />  <!-- measured from model -->
-  <property name="elbow_offset" value="-0.1719" /> <!-- measured from model -->       
+  <property name="elbow_offset" value="-0.1719" /> <!-- measured from model -->
 
   <!-- link lengths used in model -->
   <property name="shoulder_height" value="${ur10_d1}" />
@@ -64,7 +64,7 @@
 
 
   <xacro:macro name="ur10_robot" params="prefix joint_limited">
-  
+
     <link name="${prefix}base_link" >
       <visual>
         <geometry>
@@ -97,7 +97,7 @@
       </xacro:if>
       <dynamics damping="0.0" friction="0.0"/>
     </joint>
-    
+
     <link name="${prefix}shoulder_link">
       <visual>
         <geometry>
@@ -120,7 +120,7 @@
     <joint name="${prefix}shoulder_lift_joint" type="revolute">
       <parent link="${prefix}shoulder_link" />
       <child link = "${prefix}upper_arm_link" />
-      <origin xyz="0.0 ${shoulder_offset} 0.0" rpy="0.0 ${pi / 2.0} 0.0" />    
+      <origin xyz="0.0 ${shoulder_offset} 0.0" rpy="0.0 ${pi / 2.0} 0.0" />
       <axis xyz="0 1 0" />
       <xacro:unless value="${joint_limited}">
         <limit lower="${-2.0 * pi}" upper="${2.0 * pi}" effort="330.0" velocity="2.16"/>
@@ -281,7 +281,7 @@
         <origin xyz="0.0 0.0 0.0" rpy="0 0 0" />
       </xacro:cylinder_inertial>
     </link>
-    
+
     <joint name="${prefix}ee_fixed_joint" type="fixed">
       <parent link="${prefix}wrist_3_link" />
       <child link = "${prefix}ee_link" />
@@ -299,6 +299,6 @@
 
     <xacro:ur10_arm_transmission prefix="${prefix}" />
     <xacro:ur10_arm_gazebo prefix="${prefix}" />
-  
+
   </xacro:macro>
 </robot>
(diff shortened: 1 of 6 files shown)


gavanderhoorn:
See ros-industrial/ros_industrial_issues#24 for a rationale for these frames and #49 and #95 for the `universal_robot` related issues.

abubeck:
Felix tested your transformation and tree-like structure and it works. Both Felix and myself prefer your structure over mine in the local branch.
The link and joint names you used in this PR are very generic. What do you think about naming "base" to "ur_base_link" and "tool" to "ur_ee_link" and the joints in a corresponding way?

gavanderhoorn:
I understand why you'd say that, but the idea behind these frame names is to have standardised names and semantics (and the referenced issue on ros_industrial_issues contains some rationale). So I'd really like to keep them as they are.

gavanderhoorn:
Ping?

gavanderhoorn:
#80 is also related. The `frame_id` in the `WrenchStamped` header could be set to `base` after that frame has been added to the urdf, which would correspond to what the URScript manual states.

The code should probably take the `~prefix` parameter into account as well.

ur_description/urdf/ur10.urdf.xacro
```diff
@@ -299,6 +299,26 @@
 
     <xacro:ur10_arm_transmission prefix="${prefix}" />
     <xacro:ur10_arm_gazebo prefix="${prefix}" />
-  
+
+    <!-- ROS base_link to UR 'Base' Coordinates transform -->
+    <link name="${prefix}base"/>
+    <joint name="${prefix}base_link-base" type="fixed">
```
fmessmer: can we at least have some kind o `_joint` in the joint name?

ur_description/urdf/ur10.urdf.xacro
```diff
@@ -299,6 +299,26 @@
...
+      <!-- NOTE: this rotation is only needed as long as base_link itself is
+                 not corrected wrt the real robot (ie: rotated over 180
+                 degrees)
+      -->
```
fmessmer, line 309: do we need this comment?
I think not as the `base_link` will not be corrected due to backwards compatibility...

ur_description/urdf/ur10.urdf.xacro
```diff
@@ -299,6 +299,26 @@
...
+      <origin xyz="0 0 0" rpy="0 0 ${-pi}"/>
+      <parent link="${prefix}base_link"/>
+      <child link="${prefix}base"/>
+    </joint>
+
+    <!-- Frame coincident with all-zeros TCP on UR controller -->
+    <link name="${prefix}tool0"/>
+    <joint name="${prefix}wrist_3_link-tool0" type="fixed">
```
fmessmer: can we at least have some kind o `_joint` in the joint name?

ur_description/urdf/ur10_joint_limited_robot.urdf
```diff
@@ -282,6 +282,22 @@
       <mechanicalReduction>1</mechanicalReduction>
     </actuator>
   </transmission>
+  <link name="base"/>
+  <joint name="base_link-base" type="fixed">
```
ipa-fxm: can we at least have some kind o ```_joint``` in the joint name?

ur_description/urdf/ur10_joint_limited_robot.urdf
```diff
@@ -282,6 +282,22 @@
...
+    <!-- NOTE: this rotation is only needed as long as base_link itself is
+                 not corrected wrt the real robot (ie: rotated over 180
+                 degrees)
+      -->
+    <origin rpy="0 0 -3.14159265" xyz="0 0 0"/>
+    <parent link="base_link"/>
+    <child link="base"/>
+  </joint>
+  <link name="tool0"/>
+  <joint name="wrist_3_link-tool0" type="fixed">
```
ipa-fxm: can we at least have some kind o ```_joint``` in the joint name?

ur_description/urdf/ur10_joint_limited_robot.urdf
```diff
@@ -282,6 +282,22 @@
       <mechanicalReduction>1</mechanicalReduction>
     </actuator>
   </transmission>
+  <link name="base"/>
+  <joint name="base_link-base" type="fixed">
+    <!-- NOTE: this rotation is only needed as long as base_link itself is
+                 not corrected wrt the real robot (ie: rotated over 180
+                 degrees)
+      -->
```
ipa-fxm: do we need this comment?
I think not as the ```base_link``` will not be corrected due to backwards compatibility...

ur_description/urdf/ur5.urdf.xacro
```diff
@@ -315,5 +315,25 @@
     <xacro:ur5_arm_transmission prefix="${prefix}" />
     <xacro:ur5_arm_gazebo prefix="${prefix}" />
 
+    <!-- ROS base_link to UR 'Base' Coordinates transform -->
+    <link name="${prefix}base"/>
+    <joint name="${prefix}base_link-base" type="fixed">
```
fmessmer: can we at least have some kind o `_joint` in the joint name?

ur_description/urdf/ur5.urdf.xacro
```diff
@@ -315,5 +315,25 @@
...
+                 degrees)
+      -->
+      <origin xyz="0 0 0" rpy="0 0 ${-pi}"/>
+      <parent link="${prefix}base_link"/>
+      <child link="${prefix}base"/>
+    </joint>
+
+    <!-- Frame coincident with all-zeros TCP on UR controller -->
+    <link name="${prefix}tool0"/>
+    <joint name="${prefix}wrist_3_link-tool0" type="fixed">
```
fmessmer: can we at least have some kind o `_joint` in the joint name?

ur_description/urdf/ur5.urdf.xacro
```diff
@@ -315,5 +315,25 @@
     <xacro:ur5_arm_transmission prefix="${prefix}" />
     <xacro:ur5_arm_gazebo prefix="${prefix}" />
 
+    <!-- ROS base_link to UR 'Base' Coordinates transform -->
+    <link name="${prefix}base"/>
+    <joint name="${prefix}base_link-base" type="fixed">
+      <!-- NOTE: this rotation is only needed as long as base_link itself is
+                 not corrected wrt the real robot (ie: rotated over 180
+                 degrees)
+      -->
```
fmessmer, line 324: do we need this comment?
I think not as the `base_link` will not be corrected due to backwards compatibility...

gavanderhoorn:
@ipa-fxm: the urdfs are generated from the xacros, so one comment per xacro would've probably sufficed ..

ur_description/urdf/ur10.urdf.xacro
gavanderhoorn: If you feel strongly about it, I can certainly add that.

ur_description/urdf/ur10.urdf.xacro
gavanderhoorn, line 309: Yes, I think we do. First, it is a rationale for why there is a rotation between `base` and `base_link` in the first place, and secondly, I do believe we should fix-up `base_link`, but do so using bw-compatibility respecting release practices (ie: tick-tock).

gavanderhoorn:
I've added the requested suffix. Looking at the other joints (such as `ee_fixed_joint`), it seemed `_fixed_joint` instead of just `_joint` was more appropriate.

gavanderhoorn:
Anything else you require before this can be merged?

fmessmer:
Sorry @gavanderhoorn, I'm still waiting for a reply from @abubeck!
@abubeck: Could you please let us know whether your feature works with this URDF extension?

abubeck:
I still don't like the link names but if this is according to the ros-i spec I'm fine with it.
I think we can merge this now.

gavanderhoorn:
@abubeck: the main idea was to be able to do ROS -> robot Cartesian transformations regardless of which urdf (robot model) is configured in the cell. For that, we'd need some standardised frame names, which cannot include any references to a mfg or model (as that would make it robot specific again).

`base_link` is already 'taken' by ROS, and its semantics are quite different: for most urdfs, it is the link that is the root of the frame/link tree, and there is no requirement for that root to be at the 'origin' of a robot-local coordinate system. Obviously, relocating all `base_link`s in all ROS-I urdfs/xacros was out of the question, so we had to introduce another frame. That became `base`.

Same story for `tool0`, although here the idea is that the frame should be coincident with an _all-zeros toolframe_. There are (industrial) robots for which an all-zeros toolframe is _not_ coincident with the (mathematical) location of the last link, so we could not call the frame `ee_link` or similar (and: 'ee' == 'end effector', which is really not the same as a toolframe). Also: the `tool0` frame typically has no geometry associated with it (it is only a transform), while for many robots, something like an `ee_link` could mean the flange, or some other physical attachment point on the arm.

Finally, users are not supposed to update `tool0` in case they configure a different TCP, but are supposed to _attach_ their tool-urdfs to `tool0`. That way, the robot's urdf can always stay the same, and tools will always have proper collision geometry associated with their TCP frames.

For the UR currently, the `base_link` frame _is_ in the correct _location_, but it doesn't have the correct _orientation_, hence the rotation. The `ee_link` is a bit special: it has the same issue, but is somewhat similar already to `tool0` as it is only a frame, which makes the addition of another a bit awkward. I saw no other way though without breaking backwards compatibility.

abubeck:
Thanks for your extensive explanation. You convinced me wrt. the robot independent naming.
As I said I'm fine with merging. And I will do so.

gavanderhoorn:
Thanks. If I'm not mistaken, solving #80 should now also be trivial. I have no time to test that any time soon though, so if someone else would be willing to do that, that would be appreciated.

PR #215, opened 2015-09-10
Moved SetIO FUN constants from driver.py to relevant srv file

ThomasTimm:
for better interaction from other nodes


fmessmer:
Thanks @ThomasTimm...this definitely makes sense!
...and sorry for letting this hang for such a long time!

PR #219, opened 2015-09-24
Added Query state service

rohitkalaskar:
It's a service to query the state of the arm at any future time. Just like in Joint Trajectory Controller http://wiki.ros.org/joint_trajectory_controller#Services
I found it helpful for online trajectory replacement. Please let me know if there's a better way to get the future states while the arm is moving.

### DIFF
--- a/ur_driver/src/ur_driver/driver.py
+++ b/ur_driver/src/ur_driver/driver.py
@@ -22,6 +22,7 @@ from ur_driver.cfg import URDriverConfig
 from ur_driver.deserialize import RobotState, RobotMode
 from ur_driver.deserializeRT import RobotStateRT
 
+from control_msgs.srv import *
 from ur_msgs.srv import SetPayload, SetIO
 from ur_msgs.msg import *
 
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
I know this is an old PR (apologies for that), and it's completely fine if you don't respond, but I just wanted to point out that the proposed implementation of the `query_state` service callback does not actually return the current state of the robot, but an open-loop approximation based on where the robot *should be* in the current trajectory execution according to the time difference between `start_of_trajectory` and `time_now`.

Given the fact that the robot could be stopped, slower than expected or in any other way not at the position we expect it to be, I feel this is not a good way to implement something called `query_state`, which seems to imply that it returns the actual current state of the robot.

Although we don't officially accept any feature enhancements to `ur_driver` (it's deprecated in favour of `ur_modern_driver`), I would merge this given its age and if/when it's updated to return the actual current state of the robot (perhaps the joint states coming in from the controller could be used).

gavanderhoorn:
I'm going to close this.

Both because of inactivity (for which I'm partly to blame as well), but also because this PR does not actually implement what it should and would need changes to get it there.

As `ur_driver` is deprecated and will not receive any further development, if this functionality is still desired by anyone they should consider submitting a PR to the replacement repositories.

gavanderhoorn:
Nevertheless, thanks a ton @rohitkalaskar for making the effort to implement this and to provide a PR. That is always much appreciated.

PR #226, opened 2016-01-13
fix inertia for ur5

k-okada:

PR #227, opened 2016-01-20
enabled to use effort controller with 'use_effort_controller'

k-okada:
This reverts #161, I understand the motivation listed on #161, but position control will produce unstable motion when the robot contacts with object/environment, this is more stable for such situation.

### DIFF
--- a/ur_description/urdf/ur.transmission.xacro
+++ b/ur_description/urdf/ur.transmission.xacro
@@ -6,7 +6,12 @@
     <transmission name="${prefix}shoulder_pan_trans">
       <type>transmission_interface/SimpleTransmission</type>
       <joint name="${prefix}shoulder_pan_joint">
-        <hardwareInterface>PositionJointInterface</hardwareInterface>
+        <xacro:if value="$(arg use_position_interface)">
+          <hardwareInterface>PositionJointInterface</hardwareInterface>
+        </xacro:if>
+        <xacro:if value="$(arg use_effort_interface)">
+          <hardwareInterface>EffortJointInterface</hardwareInterface>
+        </xacro:if>
       </joint>
       <actuator name="${prefix}shoulder_pan_motor">
         <mechanicalReduction>1</mechanicalReduction>
@@ -16,7 +21,12 @@
     <transmission name="${prefix}shoulder_lift_trans">
       <type>transmission_interface/SimpleTransmission</type>
       <joint name="${prefix}shoulder_lift_joint">
-        <hardwareInterface>PositionJointInterface</hardwareInterface>
+        <xacro:if value="$(arg use_position_interface)">
+          <hardwareInterface>PositionJointInterface</hardwareInterface>
+        </xacro:if>
+        <xacro:if value="$(arg use_effort_interface)">
+          <hardwareInterface>EffortJointInterface</hardwareInterface>
+        </xacro:if>
       </joint>
       <actuator name="${prefix}shoulder_lift_motor">
         <mechanicalReduction>1</mechanicalReduction>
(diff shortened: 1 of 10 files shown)


k-okada:
Testing with this script;

```
#!/usr/bin/env python

import rospy
import actionlib
from control_msgs.msg import FollowJointTrajectoryAction, FollowJointTrajectoryActionGoal, FollowJointTrajectoryGoal
from trajectory_msgs.msg import JointTrajectory, JointTrajectoryPoint

rospy.init_node('send_motion')
rospy.sleep(1)

client = actionlib.SimpleActionClient('arm_controller/follow_joint_trajectory', FollowJointTrajectoryAction)
client.wait_for_server()

msg = FollowJointTrajectoryActionGoal()
msg.goal_id.id = 'id'
msg.goal_id.stamp = rospy.get_rostime()
msg.goal.trajectory.header.stamp = rospy.Time.now() + rospy.Duration(0.2)
msg.goal.trajectory.joint_names = ['shoulder_pan_joint', 'shoulder_lift_joint', 'elbow_joint', 'wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint']
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,0,0,0,0,0], time_from_start = rospy.Duration(1)))
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,-1.5704,0,0,0,0], time_from_start = rospy.Duration(2)))
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,-1.5704,0,0,0,0], time_from_start = rospy.Duration(3)))
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,-0.5236,1.0472,-0.5326,0,0], time_from_start = rospy.Duration(4)))
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,-0.5236,1.0472,-0.5326,0,0], time_from_start = rospy.Duration(5)))
msg.goal.trajectory.points.append(JointTrajectoryPoint(positions=[0,0,1.5704,0,0,0], time_from_start = rospy.Duration(6)))
client.send_goal(msg.goal)
client.wait_for_result()

```

k-okada:
i have changed my mind, added xacro interface to use effort_interface if you like, default behavior has not been changed. 

```
+  <xacro:arg name="use_position_interface" default="true" />
+  <xacro:arg name="use_effort_interface" default="false" />
```

shaun-edwards:
@k-okada, thanks for this contribution.  Sorry it took us so long to respond.  We aim to be better from here on out.  
@ThomasTimm, can you look at this one.  I don't have experience with the `HardwareInterface` functionality.

ThomasTimm:
To my knowledge it is not possible to control the joint effort on the UR and as such, I'm not sure it makes much sense to implement a `EffortJointInterface` in the URDF, seeing that the UR controller does not expose such an interface. Thus users might find themselves spending valuable time developing a solution that works in simulation, only to find that it can't be used on a real robot.

It might be possible to hack a Cartesian-based effort interface together, but I am not sure how well that would work. Also, I haven't heard of a `EffortCartesionInterface` in ROS yet, just as I haven't heard of UR planning to expose a joint effort interface. But my sources for that isn't any better than what is publicly available.

On the other hand, implementing a `VelocityJointInterface` in the URDF would make a lot of sense as that is supported by the controller and already implemented in the ur_modern_driver

gavanderhoorn:
The way this change is implemented would seem to be fully bw-compatible, so we could potentially merge this.

The PR does however seem to include e586420c2d85b907351c4d9faf68af531dbdbd91, which is part of #226. It would be good if we could keep these changes separate.

@k-okada: is this something you think would be possible? Let me know, otherwise I'll create a version of this PR without e586420c2d85b907351c4d9faf68af531dbdbd91.

gavanderhoorn:
We could probably simplify the PR quite a bit by making use of Jade+ xacro with default arguments (and then making the `hardwareInterface` value a property). That would probably allow us to avoid the many `if-else` repeats.

k-okada:
@gavanderhoorn yes, using effort controller and changing inertial values can be separated. But I think effort controller requires better inertial modeling... so  I think it better to merge #226 first.

nag92:
Will the effort controller work on real hardware?

gavanderhoorn:
@nag92: no, this is just for simulation.

UR does not expose a joint torque control interface.

gavanderhoorn:
Related: #521.

fmauch:
Closing this in favor of #619

I guess, targeting indigo-devel is not relevant anymore...

Issue #234, opened 2016-02-19
suggestion: use args for ur_description


ThomasTimm:
Right now we have two robot.urdf.xacro files for each robot; one with limited joint movements and another without.

As Xacro has support for arg (just like launch files), I think it would make more sense to just have one file, and then pass the limited arg from the launch file to the xacro file.

The same could be applied for the prefix tag, instead of ppl having to make a copy of the robot.urdf.xacro file if they are using two arm and thus needs to specify a prefix.

ThomasTimm:
The part concerning the prefix will only be possible once we move to jade, as proper conditional evaluation is needed.

gavanderhoorn:
Yeah. And that is exactly why the current implementation is as it is right now ..

ThomasTimm:
The issue is with handling empty strings; for some reason xacro is really bad at handling that. The 'limited' parameter will always be either true or false, so that can still be changed now.

Also, I guess ppl would make a new robot.urdf.xacro file anyway if they were using two arms - otherwise the two arms would be located on top of each other. So I guess the latter part about making the prefix tag an arguement is irrelevant.

gavanderhoorn:
Related: #144.

Yaml with substitution args or even the new [yaml support in xacro](http://wiki.ros.org/xacro#YAML_support) could be approaches to consider here.

Configuring the xacro macro from a yaml file completely would have my preference.

dhled:
Didn't finish it yet. Will try to push it this week.

dhled:
Proposition of solution: https://github.com/ros-industrial/universal_robot/pull/371

PR #235, opened 2016-02-19
Changed ur_description files to use arguments instead of maintaining …

ThomasTimm:
…two files for limited/non-limited joints. Also made it possible to specify joint prefixes with an argument.

This meets the suggestion in #234


ThomasTimm:
Currently this has some issues when not using any prefix in versions prior to jade. So don't merge just yet.

PR #236, opened 2016-02-20
add missing dependency for moveit_simple_controller_manager

### LINKED ISSUE #225: Missing dependency to run UR robots Gazebo simulation (ur_gazebo)
The `moveit_simple_controller_manager` need to be installed, as identified here:
https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/f4jF8s8_U40

fmessmer:
`moveit_plugins` brings in the dependency to `moveit_simple_controller_manager` (see [rosdistro](https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml#L5236-L5246))

fixes #225

### DIFF
--- a/ur10_moveit_config/package.xml
+++ b/ur10_moveit_config/package.xml
@@ -17,6 +17,7 @@
   <run_depend>moveit_ros_move_group</run_depend>
   <run_depend>moveit_planners_ompl</run_depend>
   <run_depend>moveit_ros_visualization</run_depend>
+  <run_depend>moveit_plugins</run_depend>
   <run_depend>joint_state_publisher</run_depend>
   <run_depend>robot_state_publisher</run_depend>
   <run_depend>xacro</run_depend>
--- a/ur3_moveit_config/package.xml
+++ b/ur3_moveit_config/package.xml
@@ -17,6 +17,7 @@
   <run_depend>moveit_ros_move_group</run_depend>
   <run_depend>moveit_planners_ompl</run_depend>
   <run_depend>moveit_ros_visualization</run_depend>
+  <run_depend>moveit_plugins</run_depend>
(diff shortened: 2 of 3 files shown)


gavanderhoorn:
`moveit_plugins` is a metapackage, and iirc using that for run or build dependencies is considered _not done_ (depending on transitive dependencies for your direct dependencies is not recommended).

Would it perhaps make sense to just add `moveit_simple_controller_manager` directly?

fmessmer:
```
$ rosdep install --from-paths src -i
executing command [sudo -H apt-get install ros-indigo-moveit-plugins]
[sudo] password for fxm: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following extra packages will be installed:
  ros-indigo-moveit-ros-control-interface
  ros-indigo-moveit-simple-controller-manager
The following NEW packages will be installed:
  ros-indigo-moveit-plugins ros-indigo-moveit-ros-control-interface
  ros-indigo-moveit-simple-controller-manager
0 upgraded, 3 newly installed, 0 to remove and 5 not upgraded.
Need to get 0 B/298 kB of archives.
After this operation, 1,646 kB of additional disk space will be used.
Do you want to continue? [Y/n]
```

gavanderhoorn:
I'm not saying it won't work, just that it is not recommended.

gavanderhoorn:
From [REP-127 - <metapackage/>](http://www.ros.org/reps/rep-0127.html#metapackage) fi:

>  Metapackages may not install any code or other files, although package.xml does get installed automatically. They can depend on other metapackages, if desired, but regular catkin packages cannot.

and:

> Catkin packages must depend directly on the packages they use, not on any metapackages.

PR #237, opened 2016-04-01
Updates from latest release

fmessmer:
https://github.com/ros/rosdistro/pull/10950


shaun-edwards:
@ipa-fxm, it looks like these are build/release artifacts.  If that is the case, then there isn't a need to get them reviewed (this is the rare exception to the review rules - probably unwritten).

PR #245, opened 2016-06-07
Adding joint limit variables in .xacro.urdf files

philip-long:
I added joint limit variables in the ur_description/urdf/ur10.urdf.xacro
 file, to be able to change the limits of each joint quickly. 

From README.md:  
"As MoveIt! seems to have difficulties with finding plans for the UR with full joint limits [-2pi, 2pi], there is a joint_limited version using joint limits restricted to [-pi,pi]. In order to use this joint limited version, simply use the launch file arguments 'limited', i.e.: "

One of our robots operates in a restricted workspace between for example, [pi/2, 3*pi/2] for shoulder_pan_limit_joint, so I think its easier and clearer to change in

 ur_description/urdf/ur10_joint_limited_robot.urdf.xacro

This limits are actuated when joint_limited:=true by default -pi,pi

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -63,7 +63,13 @@
   </xacro:macro>
 
 
-  <xacro:macro name="ur10_robot" params="prefix joint_limited">
+  <xacro:macro name="ur10_robot" params="prefix joint_limited
+		 shoulder_pan_lower_limit    shoulder_pan_upper_limit
+		 shoulder_lift_lower_limit    shoulder_lift_upper_limit
+		 elbow_joint_lower_limit    elbow_joint_upper_limit
+		 wrist_1_lower_limit    wrist_1_upper_limit
+		 wrist_2_lower_limit    wrist_2_upper_limit
+		 wrist_3_lower_limit    wrist_3_upper_limit">
 
     <link name="${prefix}base_link" >
       <visual>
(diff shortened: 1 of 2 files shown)


miguelprada:
@philip-long I think it would be nice if you add the default values (i.e. -pi or pi) to the joint limits in the xacro macro. See [this](http://wiki.ros.org/xacro#Default_parameters).

gavanderhoorn:
A bit late perhaps, but the changes in this PR seem to be asymmetric: if `joint_limited != true`, the limits from the macro arguments are not used.

That is at least counter intuitive and should probably be fixed.

ThomasTimm:
I am not sure I understand your concern... Are you considering the case where a user specifies that the robot should not be limited but nevertheless specifies limits?

gavanderhoorn:
Well, it's a bit of a stretch maybe, but before this change:
- non `joint_limited`: `-2pi, 2pi`.
- `joint_limited`: `-pi, pi`

After this change, this has become:
- non `joint_limited`: `-2pi, 2pi`.
- `joint_limited`: `-pi, pi` _or whatever the user provides_

There is some asymmetry there, and the semantics of the two options have also changed.

Perhaps the best course of action would be to just add some hard to miss comments that clearly explain what will happen in both cases.

ThomasTimm:
I am sorry, but I still don't see the issue..? :]
If the user uses the non-joint_limited, the limit will be [-2-;2pi].
If the user uses the joint_limited, the limit will be whatever limit is specified by the user, or [-pi;pi] if the user doesn't change anything. Is this not the most desirable solution?

PR #249, opened 2016-06-23
Moveit planning execution launch files now set controller_joint_names parameter

BrettHemes-3M:
Per the ROS-Industrial suggested standard, the controller_joint_names parameter is now set to the (non-standard) joint names:
shoulder_pan_joint
shoulder_lift_joint
elbow_joint
wrist_1_joint
wrist_2_joint
wrist_3_joint

See Section 2 [here](http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot).

### DIFF
--- a/ur10_moveit_config/config/joint_names_ur10.yaml
+++ b/ur10_moveit_config/config/joint_names_ur10.yaml
@@ -1,0 +1,1 @@
+controller_joint_names: ['shoulder_pan_joint', 'shoulder_lift_joint', 'elbow_joint', 'wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint']
--- a/ur10_moveit_config/launch/ur10_moveit_planning_execution.launch
+++ b/ur10_moveit_config/launch/ur10_moveit_planning_execution.launch
@@ -2,7 +2,9 @@
   <arg name="sim" default="false" />
   <arg name="limited" default="false"/>
   <arg name="debug" default="false" />
-    
+
+  <rosparam command="load" file="$(find ur10_moveit_config)/config/joint_names_ur10.yaml"/>
+
   <!-- Remap follow_joint_trajectory -->
   <remap if="$(arg sim)" from="/follow_joint_trajectory" to="/arm_controller/follow_joint_trajectory"/>
   
--- a/ur3_moveit_config/config/joint_names_ur3.yaml
+++ b/ur3_moveit_config/config/joint_names_ur3.yaml
@@ -1,0 +1,1 @@
+controller_joint_names: ['shoulder_pan_joint', 'shoulder_lift_joint', 'elbow_joint', 'wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint']
--- a/ur3_moveit_config/launch/ur3_moveit_planning_execution.launch
+++ b/ur3_moveit_config/launch/ur3_moveit_planning_execution.launch
@@ -2,7 +2,9 @@
   <arg name="sim" default="false" />
   <arg name="limited" default="false"/>
   <arg name="debug" default="false" />
-    
+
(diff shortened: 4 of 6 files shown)


shaun-edwards:
@BrettHemes, thanks for the contribution.  I restarted the TravisCI job and things seem OK now.  

Do you have a specific use for the joint names parameter?  Is this related to your [dev email](https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/4y08fBnmWPg)?

BrettHemes-3M:
I am using the ur_modern_driver and was putting plans together for the
follow joint trajectory action and didn't know how to get the proper names
to populate the JointTrajectory names field other than hard coding them.
Thus my email.  I have since figured out better how to use moveit so I
don't really need this parameter.  I figured I'd add it, thinking it was
overlooked but that seems not to be the case.

shaun-edwards:
@BrettHemes, the `controller_joint_names` parameter is used by the generic `industrial_robot_client` to remap joints to the controller order.  Since the `universal_robot` driver is very specific, and the controller only ever controls a single robot, this flexibility isn't needed.  

I don't think the parameter is needed in general, but I wanted to make sure I wasn't missing a use case.

BrettHemes-3M:
Nope; you can disregard the PR.

FYI, in the Descartes tutorial code, this parameter is used to get the
joint names... Trying the same for the UR robots is what got me initially
confused.

shaun-edwards:
I guess that is a use case.  When not using MoveIt, how does one figure out the joint names?  Descartes wraps MoveIt, so the joint names might not be visible.

@Jmeyer1292, any thoughts on this?

Jmeyer1292:
@shaun-edwards Both @jrgnicho and I have used this parameter in our demonstration code. I have no objection to exposing joint names to ease this process. We could even do so at the `MoveitStateAdapter` level if we're concerned about the interface changes.

shaun-edwards:
@Jmeyer1292, The problem with using the `controller_joint_names` parameter, is that only in the instance of a single arm does the joint names for the manipulator match the names Descartes should use for planning.  

Are the joint names required for planning or just generating a ROS message?

Jmeyer1292:
@shaun-edwards I absolutely agree. It's just used for generating ROS messages. I'm game for either adding a joint names accessors to the robot model base, or just exposing the JointModelGroup at the specific MoveIt implementation.

jrgnicho:
@shaun-edwards The [industrial moveit tutorial](https://github.com/ros-industrial/industrial_training/blob/indigo-devel/training/ref/demo_descartes/src/plan_and_run/src/demo_application.cpp#L209) only uses that parameter to populate the joint names of a `trajectory_msgs::JointTrajectory` however, the same could be achieved by querying the joint names directly from the JointModelGroup class.  In fact, that's the way I should've done it when I wrote the tutorial since a RobotModel instance is already created in the demo program.

gavanderhoorn:
@BrettHemes: I'm going to close this PR as I believe that the pkgs in this repository are sufficiently different from 'regular' ros-i pkgs that introducing this parameter seems strange. As @shaun-edwards already mentioned, `controller_joint_names` is really intended to be used by the mapping logic in `industrial_robot_client`, and should perhaps not be used by other code.

Thanks for putting this PR together though: :+1:

gavanderhoorn:
Btw: I'm not sure what @jrgnicho describes is still an issue, but I agree that code should probably get the joint names for a specific group from the `RobotModel` directly instead of using the `controller_joint_names` parameter if possible.

@jrgnicho: I've tried looking for the issue you describe in the current version of `industrial_training`, but can't seem to find it. If this is still a problem I'd like to open an issue on the `industrial_training` tracker so it can be addressed.

jrgnicho:
@gavanderhoorn  I just looked and it appears that the recent changes there have done away with that issue. The tutorial I wrote is no longer there in its original form if at all.

PR #255, opened 2016-08-03
ur_gazebo: remove run_depend on gazebo_ros_pkgs meta pkg

achim-k:
Causing build error:
`Abandoned <<< ur_gazebo                                                            [ Depends on unknown jobs: gazebo_ros_pkgs ]`

### DIFF
--- a/ur_gazebo/package.xml
+++ b/ur_gazebo/package.xml
@@ -18,7 +18,6 @@
 
   <buildtool_depend>catkin</buildtool_depend>
   <run_depend>ur_description</run_depend>
-  <run_depend>gazebo_ros_pkgs</run_depend>
   <run_depend>gazebo_ros</run_depend>
   <run_depend>gazebo_ros_control</run_depend>
   <run_depend>ros_controllers</run_depend>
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Looks like `ur_gazebo` depends on a metapackage (`gazebo_ros_pkgs`) which isn't allowed and causes `catkin_tools` to not build `ur_gazebo`. This is a known issue (?) with `catkin_tools`, but `ur_gazebo` shouldn't depend on metapackages in the first place.

The proposed change however only removes the metapackage reference. @achim-k: could you make sure that `ur_gazebo`'s dependencies are now still sufficiently documented? We'd probably need to add some other dependencies.

achim-k:
I checked all nodes that are launched and all dependencies seem to be satisfied. For completeness I added a run_depend on `controller_manager` although it should already be satisfied by depending on `gazebo_ros_control`.

gavanderhoorn:
Thanks for checking @achim-k.

Appreciated.

gavanderhoorn (approved):

gavanderhoorn:
This is merged in 0c34123c779d4076a75acf892571f8ec4aa018cc.

Thanks for contributing @achim-k, much appreciated.

PR #261, opened 2016-09-01
Don't depend on moveit_plugins metapackage

jonbinney:
Instead, depend on the individual plugin packages needed. Regular packages aren't supposed to depend on metapackages. For an explanation of why, see https://github.com/catkin/catkin_tools/issues/370#issuecomment-219157118

I've tested this on kinetic so far (https://github.com/ros-industrial/universal_robot/pull/254 is also needed to run on kinetic). If someone could test this PR on indigo, that would be great.

### DIFF
--- a/ur10_moveit_config/package.xml
+++ b/ur10_moveit_config/package.xml
@@ -17,7 +17,8 @@
   <run_depend>moveit_ros_move_group</run_depend>
   <run_depend>moveit_planners_ompl</run_depend>
   <run_depend>moveit_ros_visualization</run_depend>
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Thanks.

Similar to #255.

And actually known since #236.

jonbinney:
Is there anything blocking merging of this? Let me know if any changes are needed.

gavanderhoorn (approved):

gavanderhoorn:
I've merged this in a58f4b506ae8bd774829fb6b7ceede1c1b812d15.

Thanks for the PR @jonbinney.

PR #264, opened 2016-09-29
Add dependency on new moveit_kinematics package

jonbinney:
Otherwise ur_kinematics won't build against the kinetic-devel branch of the moveit repo. It fails with:

```
universal_robot/ur_kinematics/include/ur_kinematics/ur_moveit_plugin.h:97:74: fatal error: moveit/kdl_kinematics_plugin/chainiksolver_pos_nr_jl_mimic.hpp: No such file or directory
```

In order to build on kinetic right now, with moveit kinetic-devel branch, you will also need the fix in https://github.com/ros-industrial/universal_robot/pull/254

Can someone create a `kinetic-devel` branch that we can merge the various kinetic-specific fixes into?

EDIT: Corrected moveit branch name from "master" to "kinetic-devel"

### DIFF
--- a/ur_kinematics/CMakeLists.txt
+++ b/ur_kinematics/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(ur_kinematics)
 
 find_package(catkin REQUIRED COMPONENTS roscpp geometry_msgs moveit_core
-  moveit_ros_planning pluginlib tf_conversions)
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Thanks for this @jonbinney, but I think I'll merge #275 as it also updates the `CATKIN_DEPENDS` list in the `catkin_package(..)` invocation.

Other than that this PR is identical.

jonbinney:
Makes sense. Thanks @gavanderhoorn !

PR #267, opened 2016-10-13
Support for Xamla's rounded cap (ur5)

henningkayser:
[Xamla's 3D-printable cap](https://github.com/Xamla/xamla_egomo/tree/master/egomo_3d_printed_parts/egomo-1) for wrist 2 is rounded so that the collision free space is increased. Since several groups are using this custom piece of hardware an option to enable the corresponding visual and collision meshes would be convenient.

This commit adds an optional parameter `xamla_cap` to the macro `ur5_robot`, setting the geometry meshes of wrist 2 to those provided in [Xamla's repository](https://github.com/Xamla/xamla_egomo).


gavanderhoorn:
Hi, and thanks for the PR. We always welcome all contributions, so I appreciate the fact that you made the effort and decided to submit a PR. :beers:.

However, two concerns make me hesitant to merge this right now.

First: as a matter of policy, all our support packages are supposed to model robot kinematics and geometry as faithfully as possible, basing it only on mfg specs and CAD data. I'm sure you can appreciate why it's important that our models are accurate and correct. Including aftermarket mods would be a clear violation of that policy.

Second: I'm not against including an optional feature, especially if it's popular, but it would become problematic if we'd have to support every other mod out there. This PR introduces only a single additional argument, but scalability and maintainability is going to suffer if we have 70+ of them. Obviously I just made up that number, but my point remains: either we accept all user contributed mods, or none.

Now (as I wrote earlier) I can see value in having a central location to store these kind of mods, especially if they make working with the robot easier, safer or work-around current limitations. I'm open to suggestions on how to best manage this.

v4hn:
Hey @gavanderhoorn, thanks for looking at this so soon!

I asked @henhenhen to prepare this patch after talking to @andreaskoepf from Xamla.
The 3d-printable cap removes a really nasty self-collision that UR apparently tolerated in order to have uniform caps on all their links.

About _everyone_ who works with the ur5 for a while hits this collision at some point:
When the wrist links point "inwards" and align with the lower arm link, the last link cannot rotate.
wrist2_collision
This is pretty much a pointless restriction of the workspace.

With the cap, this collision does not occur:
wrist2_no_collision

I consider this to be an optional fix for the official hardware and not a mod,
so I'm much in favor of having support for this in the repository.

Because Xamla was the group that proposed and built the new cap, imho it is justified to leave their logo on the cap.

gavanderhoorn:
Hi @v4hn.

> About _everyone_ who works with the ur5 for a while hits this collision at some point:

The fact that a lot of users might be using it doesn't make it any less of a mod :).

> I consider this to be an optional fix for the official hardware and not as a mod, so I'm much in favor of having support for this in the repository.

I see your point, but again, I'm hesitant and mainly because I'd like to keep the macros as simple as possible, without (lots of) additional arguments if they aren't absolutely necessary.

It's unfortunate that xacro doesn't support some nice way to overlay changes or overwrite only small parts of a 'parent' model.

I'll let this soak for a bit and see if I can come up with an alternative way of supporting this without having a non-vendor part added.

gavanderhoorn:
@ThomasTimm, @shaun-edwards: thoughts?

v4hn:
> It's unfortunate that xacro doesn't support some nice way to overlay changes or overwrite only small parts of a 'parent' model.

Indeed. This makes it difficult to maintain third-party variations of descriptions.
But then I think, I don't want to see the additional complexity such a feature would introduce to xacro..

> I'll let this soak for a bit and see if I can come up with an alternative way of supporting this without having a non-vendor part added.

I don't see such an alternative.
If someone would refactor the `ur5_robot` macro to use helper-macros for the different parts, we could create a new xacro file that could make use of these helpers too. But even then it would boil down to cloning the current package and maintaining the clone just to change this one link mesh.

The plus of this particular non-vendor part is that it can be 3d printed. :)

gavanderhoorn:
After discussing this internally, we don't see a way to add support for this without running into the 'add special cases for everything' issue.

> If someone would refactor the `ur5_robot` macro to use helper-macros for the different parts,

Yes, that would be an option. I have some ideas on how to do this, and I've noted the inclusion of something like this in gavanderhoorn/universal_robot#1.

I'm going to close this for now, but do know that it's on our radar.

v4hn:
Fair enough.

> > If someone would refactor the ur5_robot macro to use helper-macros for the different parts,
> 
> Yes, that would be an option.

Would you welcome a refactoring of the respective xacro files in the kinetic-devel branch if someone finds the time to do that?

gavanderhoorn:
I don't know whether there is still interest in this @v4hn, but #371 is trying to implement an approach that would make using custom meshes possible.

Issue #265, opened 2016-10-11
Elbow joint self-collisions break path planning when full joint space is used on UR5


scottpaulin:
When the elbow joint is close to Pi or -Pi  the UR5 will be in self-collision, no matter what positions the other joints are in:
![ur5_self_collision](https://cloud.githubusercontent.com/assets/9565877/19257027/c9e257de-8fc8-11e6-993d-f82d9191af53.png)
Fig. 1: Self collision when elbow joint is near -Pi or Pi. Moving the other joints will not bring the robot out of self-collision.

This results in the collision free part of the UR5's configuration space to be disjoint:
![ur5_cfree_annotated](https://cloud.githubusercontent.com/assets/9565877/19257104/78c89150-8fc9-11e6-9863-766844d5e8e0.png)
Fig 2: Collision free part of the UR5's configuration space (CFree) when only considering self-collisions. The green circles show where CFree becomes discontinuous because of the elbow joint self collisions shown in Fig. 1. CFree is made up of 3 disjoint sets A, B, and C depending on the position of the elbow joint.

The 'disjoint-ness' shown in Fig. 2 means that it is _impossible_ to plan a path where the elbow joint position starts in one of the sets A, B, or C and finishes in another. I think this is why plans fail reasonably often when using the full joint limits of the UR5. 

This could be fixed by limiting the elbow joint to [-Pi, Pi) when full joint limits are used. The other joints will be fine to use their full [-2Pi, 2Pi) range of motion. 

I have not tested with the other UR robots, but they may have a similar issue.

gavanderhoorn:
Trying to understand this better: MoveIt will refuse/fail to plan whenever the start state puts the robot in a (self) collision. Isn't that what is happening here?

Not saying this isn't a problem, jsut trying to figure out "whose" problem it is.

ThomasTimm:
It is not the start state that is in a self collision, it is part of the path.
If the robot needs to move the elbow joint from 2.5rad to 0.36rad, the same configuration would be to have the arm at 3.5rad, so the inverse kinematics solver chooses that solution. But that requires the arm to move through itself and thus the planning fails (as I've understood it).
Although only giving it a few minutes of thoughts, I think it is a reasonable solution. It kind of makes me wonder why the elbow joint has such a large motion range, in reality it can never move outside [pi;-pi)..?

gavanderhoorn:
Ah, makes sense. Unconsciously I had added "at the start of planning" everywhere. Not sure why.

[...]

PR #268, opened 2016-11-02
Fix elbow joint limits

beatrizleon:
We want to propose this PR that will fix the joints limits of the elbow joint according with the results discussed in https://github.com/ros-industrial/universal_robot/issues/265 for the ur10 and ur5

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -162,7 +162,7 @@
       <origin xyz="0.0 ${elbow_offset} ${upper_arm_length}" rpy="0.0 0.0 0.0" />
       <axis xyz="0 1 0" />
       <xacro:unless value="${joint_limited}">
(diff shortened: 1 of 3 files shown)


v4hn:
@ipa-fxm @gavanderhoorn This issue has been raised and analyzed **a year ago**.
**Please merge&release at least** #302 to get this fixed in kinetic.
There is still no official driver available for kinetic (the rewritten driver is still pending), so I believe it is well-justified to merge this into indigo too.

Today I talked to one more PhD student who fell into this trap last year and the lab decided not to use MoveIt! after trying for some time. This is the second time I hear about something like that.
This is awful and bad advertisement for Motion Planning in ROS.

gavanderhoorn:
Not a change that CI has any tests for, so I feel it's justified to merge without waiting for Travis to complete.

gavanderhoorn:
Thank you for taking the time to fix this @beatrizleon :100:

gavanderhoorn:
And we should have merged this sooner :disappointed:

PR #272, opened 2016-11-30
New message definition for the ToolData

NikolasE:
This PR contains a new message definition for the ToolData that contains the analog inputs from the tool which where not available so far. The message contains numerical constants so that the user can understand the flags and modes easier.

### DIFF
--- a/ur_msgs/msg/ToolDataMsg.msg
+++ b/ur_msgs/msg/ToolDataMsg.msg
@@ -17,4 +17,4 @@ uint8 TOOL_BOOTLOADER_MODE = 249
 uint8 TOOL_RUNNING_MODE = 253
 uint8 TOOL_IDLE_MODE = 255
 
-uint8 tool_mode # one of TOOL_BOOTLOADER_*
+uint8 tool_mode # one of TOOL_*
--- a/ur_msgs/msg/ToolDataMsg.msg
+++ b/ur_msgs/msg/ToolDataMsg.msg
@@ -8,7 +8,7 @@ int8 analog_input_range2 # one of ANALOG_INPUT_RANGE_*
 int8 analog_input_range3 # one of ANALOG_INPUT_RANGE_*
 float64 analog_input2
 float64 analog_input3
-float32 tool_voltage_48V
+float32 tool_voltage_48v
 uint8 tool_output_voltage
 float32 tool_current
 float32 tool_temperature
--- a/ur_msgs/msg/ToolDataMsg.msg
+++ b/ur_msgs/msg/ToolDataMsg.msg
@@ -1,0 +1,20 @@
+# This data structure contains the ToolData structure
+# used by the Universal Robots controller
+
+int8 ANALOG_INPUT_RANGE_CURRENT = 0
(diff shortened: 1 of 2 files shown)


ur_msgs/msg/ToolDataMsg.msg
```diff
@@ -0,0 +1,20 @@
...
+# used by the Universal Robots controller
+
+int8 ANALOG_INPUT_RANGE_CURRENT = 0
+int8 ANALOG_INPUT_RANGE_VOLTAGE = 1
+
+int8 analog_input_range2 # one of ANALOG_INPUT_RANGE_*
+int8 analog_input_range3 # one of ANALOG_INPUT_RANGE_*
+float64 analog_input2
+float64 analog_input3
+float32 tool_voltage_48V
```
gavanderhoorn: ROS naming conventions: the `V` should be lower case.

NikolasE: fixed

NikolasE:
@gavanderhoorn: Could you approve the PR? I could then open a PR to integrate the related parsing code into the ur_modern_driver

ur_msgs/msg/ToolDataMsg.msg
```diff
@@ -0,0 +1,20 @@
...
+float32 tool_voltage_48v
+uint8 tool_output_voltage
+float32 tool_current
+float32 tool_temperature
+
+uint8 TOOL_BOOTLOADER_MODE = 249
+uint8 TOOL_RUNNING_MODE = 253
+uint8 TOOL_IDLE_MODE = 255
+
+uint8 tool_mode # one of TOOL_BOOTLOADER_*
```
gavanderhoorn: I think the comment should read: `# one of TOOL_*`?

gavanderhoorn:
A comment on this PR: thinking about it a bit, I'm actually not really a fan of adding these kind of messages. I'm rather a fan of having ROS drivers publish messages that are interpretable on their own, instead of needing another bit of information to be able to understand their contents.

In this case, I'd rather the `ur_driver` / `ur_modern_driver` publish values for the analogue inputs / outputs that are already converted (using the data in the `ToolData` packet in the binary stream coming from the controller). That way we don't shift the responsibility of doing that to consumers of the data.

I'll still merge it, but I think we should take a good look at how we want to handle these kind of cases in the future.

ur_msgs/msg/ToolDataMsg.msg
NikolasE: fixed

gavanderhoorn:
@ThomasTimm: fyi.

gavanderhoorn:
@NikolasE: I've had to update your `indigo-devel` branch to merge in some commits from #275.

If you have the time (and opportunity), it would be nice if we could squash all your commits into one, and do a rebase on top of current `indigo-devel`. That would avoid the merge commit as well.

gavanderhoorn:
@NikolasE: ignore my last comment.

I've squashed your commits and merged them in b26b61d579e892f85e325da72d49b9f93c5bcf64.

Thanks for the PR.

PR #276, opened 2016-12-29
rebased version of #261.

gavanderhoorn:


gavanderhoorn:
Going to do this in another way.

PR #275, opened 2016-12-29
Fix #274

130s:
Please see https://github.com/ros-planning/moveit/pull/247 for the change in the upstream package. 

Now I'm running a prerelease test locally on my computer for confirmation.

Issue #274, opened 2016-12-29
ur_kinematics fails to build on indigo


tfoote:
It looks like a missing include file.

```
00:08:02 In file included from /tmp/binarydeb/ros-indigo-ur-kinematics-1.1.6/src/ur_moveit_plugin.cpp:88:0:
00:08:02 /tmp/binarydeb/ros-indigo-ur-kinematics-1.1.6/include/ur_kinematics/ur_moveit_plugin.h:97:74: fatal error: moveit/kdl_kinematics_plugin/chainiksolver_pos_nr_jl_mimic.hpp: No such file or directory
00:08:02  #include <moveit/kdl_kinematics_plugin/chainiksolver_pos_nr_jl_mimic.hpp>
00:08:02     
```

http://build.ros.org/view/Ibin_uS64/job/Ibin_uS64__ur_kinematics__ubuntu_saucy_amd64__binary/41/console

130s:
Prerelease test against Indigo-Trusty succeeded.
```
$ mkdir /tmp/prerelease_ur
$ cd /tmp/prerelease_ur
$ generate_prerelease_script.py   https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml indigo default ubuntu trusty amd64  universal_robot --level 0 --custom-repo universal_robot:git:https://github.com/130s/universal_robot:fix/274   --output-dir ./
$ ./prerelease.sh 
:
# END SUBSECTION
+ echo # END SECTION
# END SECTION

Build step 8
+ [  = false ]

No test results in overlay workspace
```

130s:
Sorry, above was wrong comment. Deleted.

PR #277, opened 2016-12-29
[CI] Use industrial_ci

130s:
I don't force but recommend `industrial_ci` that is used in many other repos in `ros-industrial`.
- (With a risk of being pedantic...) `industrial_ci` is a set of CI configs hosted at [another repository](https://github.com/ros-industrial/industrial_ci) and frees you from maintaining CI configs.
- At situations like https://github.com/ros-industrial/universal_robot/pull/275#issuecomment-269578529, if you could test against both public synced and non-synced repos then you could have seen if your PR passes test for either/both repos. industrial_ci easily lets you do that as you can see in this PR.
- Running prerelease test can also help detecting issues like #274, before merging your PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,29 @@
-sudo: required
-dist: trusty
-# Force travis to use its minimal image with default Python settings
-language: generic
+sudo: required 
+dist: trusty 
+language: generic 
 compiler:
   - gcc
+notifications:
+  email:
+    on_success: always
+    on_failure: always
+#    recipients:
+#      - jane@doe
 env:
-  global:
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
@130s: do you have any idea why https://travis-ci.org/ros-industrial/universal_robot/jobs/187549302 did not fail at line 3600?

Update: I think I found it..

130s:
Close-reopen to re-trigger Travis (now that `moveit_kinematics` is released publicly, Travis should hopefully pass).

shaun-edwards:
+1....Thanks for the contribution....merging.

PR #289, opened 2017-02-24
Remove dependency on ros_controllers metapackage.

miguelprada:
As per http://www.ros.org/reps/rep-0127.html, packages are not allowed to
depend on metapackages.

### DIFF
--- a/ur_gazebo/package.xml
+++ b/ur_gazebo/package.xml
@@ -6,7 +6,7 @@
       Gazebo wrapper for the Universal UR5/10 robot arms.
   </description>
 
-  
+
   <author>Alexander Bubeck</author>
   <author email="sedwards@swri.org">Shaun Edwards</author>
   <author email="fxm@ipa.fhg.de">Felix Messmer</author>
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
I was quite sure we had removed this already, but apparently not.

Thanks for this @miguelprada.

Related: #261 and #255.

gavanderhoorn:
Will have to be cherry-picked to `indigo-devel` as well.

miguelprada:
I'm targeting `indigo-devel` with this PR. Would you rather have it targetting `kinetic-devel` instead and then cherry-pick to `indigo-devel`?

gavanderhoorn:
Yes, I was too quick with my comment: I was waiting for Travis to finish, then change the target branch to `kinetic-devel` by editing the PR.

But if you would do that, also fine.

gavanderhoorn:
Hm, changing the base here is not too nice (some merge commits get added).

Let's just cp in the other direction.

gavanderhoorn:
Thanks for this.

Much appreciated.

miguelprada:
You're welcome!

Thanks for the quick response @gavanderhoorn!

PR #290, opened 2017-02-24
[kinetic] Remove dependency on ros_controllers metapackage.

gavanderhoorn:
#289 but for `kinetic-devel`.


gavanderhoorn:
Thanks again @miguelprada!

PR #298, opened 2017-05-06
An additional link better than more reiterations: update README with ur_driver compatibility

jacknlliu:
add a **NOTE** about `ur_driver` compatibility and a few adjustments.


gavanderhoorn:
Thanks for your PR.

While it is a good idea to draw attention to compatibility information, we don't want to have this information in multiple places. If anything changes we'll have to update it in multiple places as well, which adds to our maintenance burden.

Adding an additional link to the wiki pages is a good idea. If you could update your PR to add just the top bit, we could merge it.

jacknlliu:
Is it suitable from your side?

gavanderhoorn (approved):

gavanderhoorn:
Yes. Thanks for taking the time to improve things, much appreciated! 👍

PR #303, opened 2017-06-20
Updated urdf::ModelInterface pointer type for ROS Lunar

marcoesposito1988:
urdf_model changed the shared pointer type from boost::shared_ptr to std::shared_ptr. 

This change is only applied to versions of ROS newer than Lunar. And it was the only change necessary to compile the whole stack into Lunar.


gavanderhoorn:
Seeing as we use a branch only when necessary policy for this repository, it seems appropriate to merge this change into `kinetic-devel`, instead of creating a `lunar-devel` branch immediately.

gavanderhoorn:
Thanks for the contribution @marcoesposito1988 :+1:

gavanderhoorn:
No need to bp to `indigo-devel`, as this is a Lunar only change and we include it in `kinetic-devel` only to make building on Lunar possible.

Issue #300, opened 2017-05-21
ik for UR donation


TheURMaxer:
About 2 weeks ago, I posted a problem regarding the IK of UR robots. The issue related to recovering rotation information and the IK solution in UR_kin.cpp. I mentioned that the original paper, Analytic Inverse Kinematics for the Universal Robots UR-5/UR-10 Arms by Kelsey P. Hawkins, December 7, 2013, provided the "correct" forward kinematics fomulae, but only a sketch of the inverse kinematics. Keeping the story short, I made modifications to the UR_kin.cpp to provide the matching IK to the paper by Hawkins. I hereby post my modifications in the hope they are useful to anyone else.

Regards,
Max van Rooij

`UR5_inverse_kin_new(const double T[16], double q_sols[8*6], double q6_des)
{
	double const d1 = 0.08920f; //Denavit-Hartenberg parameters for the UR5 in [m]
	double const d4 = 0.10900f;
	double const d5 = 0.09300f;
	double const d6 = 0.08200f;
	double const a2 = -0.42500f;
	double const a3 = -0.39243f;

	int num_sols = 0;

	////////////////////////////// shoulder rotate joint (q1) //////////////////////////////
	double q1[2];
	{
		double A = T[3] - d6*T[2];
		double B = T[7] - d6*T[6];
		double R = A*A + B*B;

		double arccos = arc_cosine(d4 / square_root(R)) ;
		double arctan = arc_tangent2(B, A);
		double pos = arctan + arccos;
		double neg = arctan - arccos;
		if(std::fabs(pos) < zero_thr)
		{
			pos = 0.0;
		}
		if(std::fabs(neg) < zero_thr)
		{
			neg = 0.0;
		}
		q1[0] = pos + half_pi;
		q1[1] = neg + half_pi;

	}
	////////////////////////////////////////////////////////////////////////////////

	////////////////////////////// wrist 2 joint (q5) //////////////////////////////
	double q5[2][2];
	{
		for(int i = 0;i < 2;++i)
		{
			double numer = (T[3]*std::sin(q1[i]) - T[7]*std::cos(q1[i])-d4);
			double arccos = arc_cosine(numer/d6);
			q5[i][0] = +arccos;
			q5[i][1] = -arccos;
		}
	}
	////////////////////////////////////////////////////////////////////////////////

	{
		for (int i = 0;i < 2;++i)
		{
			for (int j = 0;j < 2;++j)
			{
[...]

Issue #252, opened 2016-07-01
UR5KinematicsPlugin fails to plan in Cartesian space


CMobley7:
To Whom It May Concern,

I ran across issues ros-industrial/universal_robot#127, ros-industrial/universal_robot#155, ros-industrial/universal_robot#184,  ros-industrial/universal_robot#202, and ros-industrial/universal_robot#241 in which portions of ur_kimematic/UR5KinematicsPlugin were rewritten to allow the ur5_arm to better plan and execute.

However, the main topic was RVIZ. RVIZ to the best of my knowledge is using forward kinematics due to the fact it already knows the ending joint angles. I attempted to perform inverse kinematic planning using Moveit!’s Python interface with the code below on a Clearpath Husky equiped with a UR5.

```
# Initialize the move group for the arm
ur5_arm = moveit_commander.MoveGroupCommander("ur5_arm")
# Get the name of the end-effector link
end_effector_link = ur5_arm.get_end_effector_link()
# Set the reference frame for pose targets
 reference_frame = "/base_link"
# Set the arm reference frame accordingly
ur5_arm.set_pose_reference_frame(reference_frame)
# Allow replanning to increase the odds of a solution
ur5_arm.allow_replanning(True)
# Allow some leeway in position (meters) and orientation (radians)
ur5_arm.set_goal_position_tolerance(0.01)
ur5_arm.set_goal_orientation_tolerance(0.05)
#Move the end effecor to the x , y, z positon
#Set the target pose in the base_link frame
target_pose = PoseStamped()
target_pose.header.frame_id = reference_frame
target_pose.header.stamp = rospy.Time.now()
target_pose.pose.position.x = 1.0 #hole_pose.position.x
target_pose.pose.position.y = .1 #hole_pose.position.y
target_pose.pose.position.z = .7 #hole_pose.position.z
target_pose.pose.orientation.x = 0 #hole_pose.orientation.x
target_pose.pose.orientation.y = 0 #hole_pose.orientation.y
target_pose.pose.orientation.z = 0 #hole_pose.orientation.z
target_pose.pose.orientation.w = 1 #hole_pose.orientation.w
# Set the start state to the current state
ur5_arm.set_start_state_to_current_state()
# Set the goal pose of the end effector to the stored pose
ur5_arm.set_pose_target(target_pose, end_effector_link)
# Plan the trajectory to the goal
traj = ur5_arm.plan()
# Execute the planned trajectory
ur5_arm.execute(traj)
[...]

PR #305, opened 2017-07-20
Added function to enumerate all possible joint angles per IK solution

a-price:
As discussed in #300, there may be many possible joint configurations for a single kinematic IK solution. This patch recursively adds joints satisfying q[i][j] + 2pi and the joint limits.

Addresses #300, and possibly #276 and #252


maxgittelman:
this fork fixes a ton of issue, please take a look

gavanderhoorn:
This should be reviewed together with #239 and #316.

I won't do that before ROSCon however, and in any case would like some additional eyes on this.

a-price:
As mentioned in #316, this fix leaves quite a few solutions to sort through. When I wrote this, I was trying to get something working that would give the correct answer with a minimum of changes to the existing code. I think a better solution would be to pass the weights and seed state into the recursion, and perform a branch-and-bound-esque pruning of the solution tree as it's built. It might also be possible to sort each joint solution individually, but we'd have to be careful to avoid the error in the original implementation. Thoughts?

a-price:
Also, is kinetic-devel the correct branch to push to? The Travis CI [error](https://travis-ci.org/ros-industrial/universal_robot/builds/255857411#L2705) seemed to be unhappy with package installation.

gavanderhoorn:

#239 and #316 should be taken into account / considered as well.

mxgrey:
I suggest closing this PR in favor of #358 which combines this one with #239 and resolves the conflicts between them.

PR #319, opened 2017-09-12
Backport #315 to indigo

gavanderhoorn:
As Jade+ xacro is used on Indigo as well, remove `pi`.


gavanderhoorn:
As Jade+ xacro is not enabled by default on Indigo, this PR has the potential to break existing setups. `--inorder` is used in the `urN_upload.launch` files, but composite xacros on Indigo that include the UR base xacro might not be loaded using `--inorder`.

Therefore I'm not backporting this to Indigo.

gavanderhoorn:
Note that Jade+ xacro is fully supported on Indigo, so we might revisit this later.

PR #320, opened 2017-09-12
Backport #268 to Indigo

gavanderhoorn:
As per subject.


gavanderhoorn:
Merging straight away as this is not tested by CI.

PR #321, opened 2017-09-12
Backport #266 to Indigo.

gavanderhoorn:
As per subject.


gavanderhoorn:
Untestable by CI at the moment.

Merging as I've tested this locally with Indigo MoveIt.

PR #322, opened 2017-09-13
driver: don't use deprecated robot_state binary.

gavanderhoorn:
As per subject.


Levi-Armstrong (approved):

fmessmer (approved):
will merge for next release #333

PR #329, opened 2017-11-21
Homogenize xacro macro arguments.

miguelprada:
Joint limits for the limited version can be set using xacro arguments for the UR10
but not for the UR3 and UR5. Same lower and upper limit arguments are added to
the UR3 and UR5 xacro macros.


miguelprada:
For consistency, I also modified the `urX_joint_limited_robot.urdf.xacro` files for the UR3 and UR5 to match the one for the UR10.

fmessmer (approved):
will merge for next release #333

PR #334, opened 2017-12-22
Use lookupParam() for parameter queries inside URKinematicsPlugin

### LINKED ISSUE #256: ur_kinematics: getting private params when moveit plugin loaded from rviz/moveit_commander
When the UR moveit plugin is being loaded via the rviz motion planning plugin or via the moveit_commander, the ur moveit plugin fails to load when an arm_prefix was set:

```
[ERROR] [1470304127.233379094, 4020.674000000]: Kin chain provided in model doesn't contain standard UR joint 'shoulder_lift_joint'.
[ERROR] [1470304127.233633625, 4020.674000000]: Kinematics solver of type 'ur_kinematics/UR10KinematicsPlugin' could not be initialized for group 'ur10'
[ERROR] [1470304127.234046407, 4020.674000000]: Kinematics solver could not be instantiated for joint group ur10.
```

The reason for this is the following: When the plugin is loaded via rviz/moveit_commander, the moveit plugin is loaded in a different namespace like e.g.
- /move_group_commander_wrappers_1470304126604460729
- /rviz_pcname_3652_1661548360095152202

Since for these namespaces the `arm_prefix` param is not set, it can't be found and the default value (empty string) is chosen (see [here](https://github.com/ros-industrial/universal_robot/blob/1dafa8aa388b6c1163e5b061a83da6dfd345c501/ur_kinematics/src/ur_moveit_plugin.cpp#L326)). It then fails [here](https://github.com/ros-industrial/universal_robot/blob/1dafa8aa388b6c1163e5b061a83da6dfd345c501/ur_kinematics/src/ur_moveit_plugin.cpp#L328-L355) where the joint names from the urdf are compared with the expected once. 

Long story short: The plugin can't retrieve private params if not loaded into the default moveit namespace (usually `move_group`) where the desired params are set.

What works for me is when replacing the [line](https://github.com/ros-industrial/universal_robot/blob/1dafa8aa388b6c1163e5b061a83da6dfd345c501/ur_kinematics/src/ur_moveit_plugin.cpp#L326)

``` cpp
private_handle.param<std::string>("arm_prefix", arm_prefix_, "");
```

with the following code:

``` cpp
arm_prefix_ = "";
std::string key;
if(private_handle.searchParam("move_group/arm_prefix", key)) {
  private_handle.getParam(key, arm_prefix_);
}
```

However, I consider this an ugly fix since it makes the assumption that the desired params are set in the `move_group` namespace which is probably not always true. 

Is there a better way to fix this?

henningkayser:
This PR replaces all _param()_ calls inside `URKinematicsPlugin` with the new _lookupParam()_ function introduced [here](https://github.com/ros-planning/moveit/pull/701) and [here](https://github.com/ros-planning/moveit/pull/714).
The function queries parameters not only from the private namespace but also from the kinematics base and enables optional subgroups.

Following namespaces are searched in the given order:
1. ~/<group_name>/
2. ~/
3. robot_description_kinematics/<group_name>/
4. robot_description_kinematics/

Parameters can now be conveniently defined inside the kinematics.yaml.
In our setup we use this to fix [this issue](https://github.com/ros-industrial/universal_robot/issues/256) by just adding the parameter `arm_prefix` under the corresponding group.

@v4hn


v4hn:
Thanks for filing the request @henhenhen .

@gavanderhoorn This will resolve a long-standing bug in handling MoveIt kinematics plugins. I would really appreciate if you could merge it soon.

mintar:
+1

mintar:
What this PR does is that you can now put `arm_prefix` and other parameters directly into your `kinematics.yaml` file, instead of having to hunt down each launch file where the `kinematics.yaml` file is loaded, and put a `<param name="arm_prefix" ...>` next to it. Also, this PR aligns the parameter handling of URKinematicsPlugin with the way the standard kinematics plugins (KDL, ikfast, ...) now do it.

@gavanderhoorn Since this PR is fully backwards compatible, is there any reason not to merge it?

codebot (approved):
:+1: I just tested this in `ur_gazebo` with `moveit` and it works as expected. Because the parameter search order is different than previous, I suppose there could be changed behavior in some downstream configurations, but in general this seems like an improvement which will simplify code and usage patterns.

mintar:
Hey @gavanderhoorn , there are three approving code reviews for this PR: one by @v4hn , one by myself, one by @codebot (Morgan Quigley). Is there anything else we can do to get this merged?

If you are too busy to take care of this, would it be okay if someone else with write access (such as @ipa-nhg or @codebot) would merge this?

mintar (approved):

gavanderhoorn:
Thanks for the contribution @henhenhen.

Apologies for taking some time to merge it.

Thanks @codebot and @mintar for the reviews.

PR #338, opened 2018-01-18
ur_moveit_plugin: fix compile error with GCC6

machinekoder:
I encountered compile errors with GCC6 under Debian Stretch. Using the abstracted SharedPtr types fixes the problem.

Alternatively, we could use `auto`.


codebot (approved):
I built and tested this PR on Ubuntu 16.04 and verified that it does no have any effect. If I chased correctly through the various `#includes` and `#defines`, this becomes a no-op on `kinetic`.

I also created a Debian Stretch docker container and verified that there is indeed a build error on GCC 6.3, as reported, and that the PR fixes this build error.

So I vote to approve :+1: thanks @machinekoder !

ipa-nhg (approved):

PR #343, opened 2018-02-16
Added run_depend for xacro

hsd-dev:


codebot (approved):
:+1:  thanks @ipa-hsd

PR #349, opened 2018-04-19
Add safety limits to URDF files

miguelprada:
This PR adds optional `safety_controller` tags to all joints in the URDF models.

It adds the following arguments to the URX xacro macros:
- `safety_limits`: (defaults to `false`) needs to be set to true to insert the `safety_controller` tags. The other arguments have no effect if this is kept to `false`.
- `safety_pos_margin`: the lower/upper limits in the `safety_controller` tag are set to `XXX_lower_limit + safety_pos_margin` and `XXX_upper_limit - safety_pos_margin` for every joint. This avoids adding 12 new arguments for each macro (lower/upper limits per joint).
- `safety_k_position` : used to set `k_position` in the `safety_controller` tag. See [safety limits explanation](http://wiki.ros.org/pr2_controller_manager/safety_limits).

The default values for `safety_pos_margin` and `safety_k_position` are set such that when the current position approaches the hard position limit, the soft velocity limit approximately approaches the hard velocity limit in the opposite direction.

Note that `k_velocity` is not exposed as an argument, and is hardcoded to `0.0`, since it only makes sense in the context of effort control, which is not available for the UR robot.

Last, I believe this should not affect behavior of any existing code, in particular `ur_modern_driver`, but would enable implementing joint limit functionality further down the road.


ipa-nhg (approved):
Tested on simulation :+1:

gavanderhoorn:
Closing as #437 has rebased this.

gavanderhoorn:
Thanks for this PR already @miguelprada.

PR #356, opened 2018-06-21
Fix overlapping variable names between robot definition files

felixvd:
Uploading multiple different UR definitions currently breaks the previous robot definitions by overwriting their link lengths and other parameters. This commit makes the property names for each robot model unique so they are independent of the others.

See the before and after pictures in Gazebo below. The UR5 was the last model to be loaded in the scene definition, so the link lengths of the UR3 have become too long, and those of the UR10 too short. 

gazebo-before

gazebo-after

I don't think this change affects anything outside of these files and I have had no problems using it in our own development.

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -10,24 +10,24 @@
   <xacro:include filename="$(find ur_description)/urdf/ur.gazebo.xacro" />
 
   <!-- Inertia parameters -->
-  <xacro:property name="base_mass" value="4.0" />  <!-- This mass might be incorrect -->
-  <xacro:property name="shoulder_mass" value="7.778" />
-  <xacro:property name="upper_arm_mass" value="12.93" />
-  <xacro:property name="forearm_mass" value="3.87" />
-  <xacro:property name="wrist_1_mass" value="1.96" />
-  <xacro:property name="wrist_2_mass" value="1.96" />
-  <xacro:property name="wrist_3_mass" value="0.202" />
+  <xacro:property name="ur10_base_mass" value="4.0" />  <!-- This mass might be incorrect -->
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Thanks for the PR. This is indeed an issue.

Instead of renaming all the properties, would moving their declarations down a bit so they are *inside* the `ur10_robot` macro definition be an option?

According to [wiki/xacro - Local properties](http://wiki.ros.org/xacro#Local_properties):

> Properties and macros defined within a macro are local to this macro, i.e. not visible outside.

atomoclast:
One possible suggestion too is to namespace each robot. Either way this would be a great PR!

gavanderhoorn:
@atomoclast: moving the declarations into the macro def should automatically achieve this. The `properties` should essentially get namespaced that way.

felixvd:
I agree, that's the cleaner way. I updated the PR accordingly and took the liberty to remove the prefix from the kinematic parameters, too. I confirmed that it works in our simulation.

dhled:
Tested on simulation.

Chamango90:
Reviewed and accepted

gavanderhoorn:
Accepting and merging this *will* break any urdfs that are currently directly using any of the `xacro:property` elements, so we will have to increase the package's version to convey that fact *and* post an announcement on ROS Discourse about it.

As the constants are 'almost' internal constants, I don't expect that to have happened to often, but we can still be clear about it.

For future readers: `${namespace.property}` can be used to access properties inside namespaces. See also the `xacro` documentation on the wiki page linked to in https://github.com/ros-industrial/universal_robot/pull/356#issuecomment-400016598.

gavanderhoorn (approved):
Approving based on reviews by @ipa-led and @ipa-jfh.

Issue #233, opened 2016-02-18
Improve meshes shading


VictorLamoine:
Fix the visual meshes shading to get smooth lightning:
http://answers.ros.org/question/226831/how-do-i-make-my-meshes-appear-shaded/

Collisions meshes should not be edited.

fmessmer:
Feel free to provide a PR for this!
#94 is related to the meshes, too

PR #359, opened 2018-07-11
Update the colors for UR visual models

ipa-nhg:

Issue #295, opened 2017-03-23
Change material color of gazebo model


jacknlliu:
Could we change material color  of gazebo model to make it more like V-REP style, rather than LightGrey for every link?

The UR model in V-REP is more closer to its real color.

shaun-edwards:
Can you provide images of the current and desired robot model?

jacknlliu:
@shaun-edwards  still no available robot model

shaun-edwards:
I mean can you provide an example of the color difference (i.e. the current color in ROS/Gazebo) vs. the desired color in V-REP?

jacknlliu:
@shaun-edwards  I can't get the material color RGB value of UR model from V-REP.  I get a screenshot from V-REP to show the difference.

UR5 V-REP model screenshot:
![UR5 V-REP model screenshot](https://raw.githubusercontent.com/jacknlliu/hello-world/master/resources/ur-vrep.png)

UR5 Gazebo model screenshot:
![UR5 Gazebo model screenshot](https://raw.githubusercontent.com/jacknlliu/hello-world/master/resources/ur-gazebo.png)

gavanderhoorn:
I think this is more a difference in lighting between V-REP and RViz than a difference in material settings.

jacknlliu:
@gavanderhoorn  I don't think it's about lighting.  If you run `roslaunch gazebo_ros rubble_world.launch` , you will find the lighting is the same as UR settings, but the visual effect is different.

ipa-nhg:
Solved by https://github.com/ros-industrial/universal_robot/pull/359
![screenshot from 2018-07-11 10-27-17](https://user-images.githubusercontent.com/650568/42559813-7aecad0e-84f5-11e8-9539-e7b2d5e349ef.png)

PR #360, opened 2018-07-11
Improved visual meshes shading (Related to issue #233)

j-polden:
Related to issue #233 

Some meshes were not smoothed. These meshes were adjusted with 'smooth' + 'edge split' commands in  blender. Edge split ratios were usually around 30 Degrees (with some variance).

Other meshes had already been smoothed beforehand, however the 'edge split' command had executed to the full 180 degrees (which caused some visual blemishes). For these cases the, the 'edge split' level was reduced to more suitable levels.

Changes were made only to the visual meshes.

An example of the smoothed result.
before_after


ipa-nhg:
@j-polden 
can this PR be combined with https://github.com/ros-industrial/universal_robot/pull/359 ??
see also the issue https://github.com/ros-industrial/universal_robot/issues/295

Issue #293, opened 2017-03-11
python interface problem


MiaoLi:
When I run the test_analytical_ik.py file, it game me a error:
"ImportError: No module named ur_kin_py"
I have tried different ways to build the module, there is no test example for this.
Any idea?

gavanderhoorn:
How did you setup your workspace? Did you source the `setup.bash` after building it (just making sure)?

I must admit I've never had a need to use `ur_kin_py` so I don't have any experience troubleshooting it.

That said this looks like a `PYTHONPATH` issue.

MiaoLi:
Finally, I made it work.
It is not the workspace problem. We need to change the CMAKELIST to make it work, also the PYTHONPATH as mentioned above.
I would suggest to either remove this file or add some tutorial doc.

gavanderhoorn:
Could you tell us what you changed?

MiaoLi:
I add the following lines in CMakeList.txt.
PYTHON_ADD_MODULE(ur_kin_py src/ur_kin_py.cpp)
target_link_libraries(ur_kin_py ur3_kin boost_python boost_numpy ${PYTHON_LIBRARIES} ${Boost_LIBRARIES} )

Build with ccatkin_make and then copy ur_kin_py.so to the folder of test_analytical_ik.py.

atenpas:
Adding what @MiaoLi suggested to _CMakeList.txt_ in _ur_kinematics_ did **not** produce an `so` file for me but a file without extension. Renaming and copying it to the folder of _test_analytical_ik.py_ allows to run _test_analytical_ik.py_. However, this results in another bug:

`File "test_analytical_ik.py", line 39, in test_q
    print i1-3, i2-3, i3-3, i4-3, i5-3, i6-3
NameError: global name 'i1' is not defined`

See https://github.com/ros-industrial/universal_robot/blob/indigo-devel/ur_kinematics/src/ur_kinematics/test_analytical_ik.py#L38.

gavanderhoorn:
Tagging as `wrid18` and adding to the `IK` project board so this can be tracked as one of the IK related issues.

ipa-nhg:
The "ImportError" should be solved by https://github.com/ros-industrial/universal_robot/pull/364

gavanderhoorn:
Fixed by #364.

PR #364, opened 2018-07-11
Missed python module definition and setup.py script

ipa-nhg:
Related to https://github.com/ros-industrial/universal_robot/issues/293


ur_kinematics/src/ur_kinematics/__init__.py
```diff
@@ -0,0 +1 @@
+from ur_kinematics import *
```
Chamango90: The import is not required. An empty `__init__.py` file is sufficient.

ipa-nhg: you are right, it is within the same package. The __init__ file can be empty

gavanderhoorn:
Suggestion: can we please rebase as much as possible, instead of merging in `kinetic-devel` all the time? Let's try to keep the commit history sane.

ipa-nhg:
@gavanderhoorn yes! sorry, I clicked the button "update branch" of the PR  to verify the commit with Travis. But you are right, it is horrible for the history of the repo...

gavanderhoorn:
No need to apologise.

Let's just keep it in mind.

Chamango90:
Reviewed and accepted

gavanderhoorn:
Thanks @ipa-nhg for fixing this.

pidipidi:
I cannot still find a python module, ur_kin_py when I run test_analytical_ik.py. Here is the message,

"ImportError: No module named ur_kin_py"

I am not an expert for the python wrapper, but I could not find any setup for ur_kin_py.cpp in CMakeList.

cambel:
@pidipidi Yeah I found the same error, after trying many things I manage to made it work.

Here is what I `did`:
First, make sure you have installed Boost.Numpy, if not you can install it from source:
```
sudo apt-get install build-essential python-setuptools libatlas-dev libatlas3gf-base

git clone https://github.com/ndarray/Boost.NumPy
cd Boost.NumPy
cmake .
make
make install
```

Then in the CMakeLists.txt file, you have to add the following

```
find_package(PythonLibs 2.7 REQUIRED)
... 
###########
## Build ##
###########
...
PYTHON_ADD_MODULE(ur_kin_py src/ur_kin_py.cpp)
target_link_libraries(ur_kin_py ur3_kin boost_python boost_numpy ${PYTHON_LIBRARIES} ${Boost_LIBRARIES} )

set_target_properties(ur_kin_py PROPERTIES 
                      LIBRARY_OUTPUT_DIRECTORY ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_PYTHON_DESTINATION})
...
#############
## Install ##
#############
install(TARGETS ur_kin_py
  ARCHIVE DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION}
  LIBRARY DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION}
)
...
```
Note that I link the library just with the ur3_kin, if you need a different robot, you can change it.

Then you can test with the `test_analytical_ik.py` but you have to also fix the import line on this file:
`from ur_kinematics.ur_kin_py import forward, inverse`
then it should work fine.

Finally, if you encounter the problem of `libboost_numpy.so: cannot open shared object file` just update your LD_LIBRARY_PATH with the location of your library, in my case /usr/local/lib64

Hope it helps!

Issue #158, opened 2015-01-29
Improving the test_move script


Jmeyer1292:
I recently wrote a tutorial on getting started with the Universal Robots and ROS-I here: http://wiki.ros.org/universal_robot/Tutorials/Getting%20Started%20with%20a%20Universal%20Robot%20and%20ROS-Industrial. Most of this was copy/paste from the readme pages here on Github.

I realize this is probably low priority, but I believe that the current test_move script executes a movement that is too aggressive. I know I was surprised by it and I know others with similar feelings. As this file is the first introduction many will have to the programmatic control interface provided by the ur_driver, I believe that it is important that we improve this file.

Specifically, I suggest a small menu with a choice of different movements and an explanation of what they are. Or perhaps simply a smaller test movement that mostly manipulates the last two joints.

gavanderhoorn:
The test script is a relic from the earliest days of the driver (see a0b1c2c9). I don't think it was ever changed since then. I fully support changing its behaviour to be more predictable and / or updating documentation. Perhaps we could even include a short video, so users know what they can expect.

gavanderhoorn:
@Jmeyer1292: have you had any time to work on this?

hsd-dev:
https://github.com/ros-industrial/universal_robot/pull/366

PR #366, opened 2018-07-11
[WIP] Improved the test_move script in ur_driver

hsd-dev:
Related to https://github.com/ros-industrial/universal_robot/issues/158

### DIFF
--- a/ur_driver/test_move.py
+++ b/ur_driver/test_move.py
@@ -1,159 +1,64 @@
 #!/usr/bin/env python
-from __future__ import print_function
-import time
-import sys
 import rospy
 import actionlib
 from control_msgs.msg import FollowJointTrajectoryAction, FollowJointTrajectoryGoal
 from trajectory_msgs.msg import JointTrajectory, JointTrajectoryPoint
-from controller_manager_msgs.utils import filter_by_type, filter_by_state
-from controller_manager_msgs.utils import ControllerLister, ControllerManagerLister
-from sensor_msgs.msg import JointState
 
-t_stamps = [2.0, 4.0, 6.0]
+joint_names = ['shoulder_pan_joint', 'shoulder_lift_joint', 'elbow_joint',
+               'wrist_1_joint', 'wrist_2_joint', 'wrist_3_joint']
 
-DEFAULT_CONTROLLER_NS = ""
+qpos_up = [0, -1.5707, 0, -1.5707, 0, 0]
+q1 = [0, -1.5707, 0.35, -1.5707, 0.35, 0]
+q2 = [0, -1.5707, -0.35, -1.5707, -0.35, 0]
 
-def get_trajectory_goal(_jnt_names, q1, q2, q3, t=t_stamps):
-  joint_num = len(_jnt_names)
+def get_trajectory_goal(joint_names, q1, q2, q3):
+  joint_num = len(jnt_names)
   g = FollowJointTrajectoryGoal()
   g.trajectory = JointTrajectory()
-  g.trajectory.joint_names = _jnt_names
+  g.trajectory.joint_names = joint_names
   g.trajectory.points = [
-    JointTrajectoryPoint(positions=q1, velocities=[0]*joint_num, time_from_start=rospy.Duration(t[0])),
-    JointTrajectoryPoint(positions=q2, velocities=[0]*joint_num, time_from_start=rospy.Duration(t[1])),
-    JointTrajectoryPoint(positions=q3, velocities=[0]*joint_num, time_from_start=rospy.Duration(t[2]))]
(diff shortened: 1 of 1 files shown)


hsd-dev:
Some changes I would like to make:
- Script assumes 6 DOF ([#L43](https://github.com/ipa-hsd/universal_robot/blob/c17922f5b664c4ec4dcd83602928164737d768cb/ur_driver/test_move.py#L43) and [#L83](https://github.com/ipa-hsd/universal_robot/blob/c17922f5b664c4ec4dcd83602928164737d768cb/ur_driver/test_move.py#L83)). Generalize that to N DOF
- Add velocity checks if possible 

This script would not be specific to UR. Would there be a better place for it?

ur_driver/test_move.py
```diff
@@ -1,106 +1,156 @@
 #!/usr/bin/env python
+from __future__ import print_function
 import time
-import roslib; roslib.load_manifest('ur_driver')
 import rospy
 import actionlib
```
ipa-nhg, line 3: missed:
`import sys`

ur_driver/test_move.py
```diff
@@ -1,106 +1,156 @@
...
-        client.wait_for_server()
-        print "Connected to server"
-        #move1()
-        move_repeated()
-        #move_disordered()
-        #move_interrupt()
-    except KeyboardInterrupt:
-        rospy.signal_shutdown("KeyboardInterrupt")
-        raise
+  _use_ros_control = False
```
ipa-nhg: @ipa-hsd `_use_ros_control`?? what does it mean?

ur_driver/test_move.py
```diff
@@ -1,106 +1,156 @@
...
+    rospy.init_node("test_move", anonymous=True, disable_signals=True)
+    print("""
+    Usage: You can use this node alongwith or without ros_control
+           In case of without, make sure to pass the namespace of FollowJointTrajectoryAction as the first argument.
+           Else, an empty namespace is assumed.
+    """)
+    
+    if _use_ros_control == True:
+      controller_ns = get_controller().name 
+    elif(len(sys.argv) > 0):
```
ipa-nhg: I think it should be:

`elif(len(sys.argv) > 1):`

ipa-nhg (changes requested):
With this minor changes. 

Working on simulation :+1:

ur_driver/test_move.py
hsd-dev: yes you are right.

ur_driver/test_move.py
hsd-dev: I wanted to read the controller's namespace in case of ros_control. Else the user passes the namespace as an arugument.

hsd-dev: removed this to avoid confusion

ur_driver/test_move.py
hsd-dev: fixed

ipa-nhg (dismissed):
Tested on simulation and working

ipa-nhg:
@gavanderhoorn please merge if its ok for you

gavanderhoorn:
@ipa-hsd @ipa-nhg: this is definitely an upgrade, but at a first glance it also looks somewhat complex.

The goal of this script is two-fold:

 1. provide an easy way to test the driver
 1. provide an example of how action clients can be used to communicate with the driver

Goal 1 is covered by this, no doubt.

I'm wondering if we're not making things too complex for goal 2 however. The script has a lot of things I would not expect in a (simple) example, such as listening for `JointState` msgs, dealing with keyboard input and it hides the typical "wait-for-server;create-goal;send-to-server;wait-for-completion" control flow somewhat.

If you agree, we could perhaps keep this file as it is now proposed and add a much simpler version that covers the second goal.

gavanderhoorn:
Independent of this: I believe it would be prudent to print a very visible warning drawing attention to the fact that the robot is going to move, and that it is going to move in a certain way. Perhaps the script should also clarify how it is going to move, and that users should make sure the robot has X and Y free space around it?

hsd-dev:
@gavanderhoorn : I have added the warning messages. 

> If you agree, we could perhaps keep this file as it is now proposed and add a much simpler version that covers the second goal.

I am not sure of the future of this repo, but maybe the new file can be added to a more generic repository where other arms can be tested as well.

ipa-nhg:
I would like merge this to the kinetic-devel branch (first the merge conflict has to be solved). The full package "ur_driver" is deprecated (even removed for new distros), then makes not sense to put here more effort, and the changes of the PR improve the current version.

@gavanderhoorn if you don't want to merge it, I suggest to close the PR, if @ipa-hsd agrees.

PR #367, opened 2018-07-11
update and unify Readme files

ipa-nhg:


gavanderhoorn:
We're going to this differently here, but we'll include the content of this PR into that.

Thanks for the initial work @ipa-nhg :+1:

PR #372, opened 2018-07-27
simplify FK equations

qqfly:
The Forward Kinematics equations can be simplified.


gavanderhoorn:
Could you give a bit more explanation of what you've done here exactly?

qqfly:
I just checked the output of my own FK equations and yours, and found that their results are the same.

You can verify the result with these MATLAB [code](https://github.com/qqfly/ur5_fk_simplification)

jacknlliu:
From the [diff split mode](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split), we can see these changes clearly. 

From the right sidebar, before computing the transform matrix `T`(until [line 55](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split#diff-1f568893a3278acd1e5108a22dd9654eR55)), we can see that the following variables are added:
- `q23`: as the previous notation, it means `q(2) + q(3)`(C style: `q[1] + q[2]`)
- `s4`, `c4`: `sin()` and `cos()` for the joint angle `q(4)` (C style: `q[3]`). (**NOTE:** actually these two variables are not needed.)
- `s23`, `c23`: `sin()` and `cos()` for the previous angle `q23`

There are many terms which only have different arithmetic signs but have the same absolute value in computing matrix `T` from the origin codes of this repo. 

And it is the main thing that this PR does, except [line 59](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split#diff-1f568893a3278acd1e5108a22dd9654eR59), [line 63](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split#diff-1f568893a3278acd1e5108a22dd9654eR63), [line 67](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split#diff-1f568893a3278acd1e5108a22dd9654eR67) in which the `c23` and `s23` are included according to the sum and difference formulas of trigonometric functions. like

sum and difference formulas of trigonometric functions

@qqfly But from the [line 67](https://github.com/ros-industrial/universal_robot/pull/372/files?utf8=%E2%9C%93&diff=split#diff-1f568893a3278acd1e5108a22dd9654eR67), we can see that the `s4` and `c4` are not needed, and this line has a more compact form via the sum and difference formulas of trigonometric functions. like
```
*T = d1 + a3*s23 + a2*s2 - d5*c234 - d6*s5*s234; T++;
```

In summary, `s4` and `c4` should be removed and the last changed line can be more simple. Other lines are OK. These changes make the codes more clearer.

ur_kinematics/src/ur_kin.cpp
```diff
@@ -46,33 +46,25 @@ namespace ur_kinematics {
...
+    *T = -s6*(s1*s5 + c234*c1*c5) - s234*c1*c6; T++;
+    *T = d6*c234*c1*s5 - a3*c23*c1 - a2*c1*c2 - d6*c5*s1 - d5*s234*c1 - d4*s1; T++;
+    *T = c1*c5 + c234*s1*s5; T++;
+    *T = -c6*(c1*s5 - c234*c5*s1) - s234*s1*s6; T++;
+    *T = s6*(c1*s5 - c234*c5*s1) - s234*c6*s1; T++;
+    *T = d6*(c1*c5 + c234*s1*s5) + d4*c1 - a3*c23*s1 - a2*c2*s1 - d5*s234*s1; T++;
+    *T = -s234*s5; T++;
+    *T = -c234*s6 - s234*c5*c6; T++;
+    *T = s234*c5*s6 - c234*c6; T++;
+    *T = d1 + a3*s23 + a2*s2 - d5*(c23*c4 - s23*s4) - d6*s5*(c23*s4 + s23*c4); T++;
```
jacknlliu, line 67: more compact form, like
```
*T = d1 + a3*s23 + a2*s2 - d5*c234 - d6*s5*s234; T++;
```

qqfly: You are right, this is more compact.

jamesn1197:
Pulled the repo, verified the .m script and got the same results.

Levi-Armstrong (approved):

PR #380, opened 2018-08-23
UR-E Series

dniewinski:
Added E-series models


gavanderhoorn:
Very nice. Thanks for this.

I'm a little unsure whether we should reuse the setup of `ur_description` for these new models though. Ideally we'd get rid of the `joint_limited` stuff, and we can probably make the urdf/xacros a little neater by using the new yaml support in `xacro`. I'd also like to separate out the Gazebo related stuff a bit more, make the xacros match the structure of the other ros-i robot support packages (`ur_3e_macro.xacro` with `ur_3e.xacro` fi) and remove the `world` links from those files (long standing issue).

I'm also wondering whether we should introduce these new variants in a new package. `ur_description` is a big package already (~25MB), and adding these variants increases that to ~72MB (!). Besides downsampling the meshes a bit, perhaps adding a `ur_e_description` (or better: `ur_e_series_support`) package would be nice.

dniewinski:
Moving it into a ur_e_description package may be a decent solution.  I agree that a lot of that refactoring should be done, but maybe for all the UR packages all at once (instead of just the ur_e ones).

We are hoping to do some UR E testing shortly and will hopefully merge any necessary changes.

I will move the changes to their own packages for now but leave the historical quirks (world, joint_limited, etc)

dniewinski:
Are the separate collision and visual meshes necessary?  The meshes between the two are the same, it's only the materials that are not present in the stl files.  Using the visual meshes as the collision meshes shouldn't affect performance, but would reduce package size

gavanderhoorn:
> Are the separate collision and visual meshes necessary? The meshes between the two are the same, it's only the materials that are not present in the stl files. Using the visual meshes as the collision meshes shouldn't affect performance, but would reduce package size

the collision quality meshes are supposed to be heavily downsampled versions of the visual ones.

Size of the meshes used for collision geometry greatly influences planning time with MoveIt, so personally I would say: yes, they are necessary.

---

Edit: just for comparison: the `visual` directory is 8.7MB, the `collision` directory is 304 KB for all meshes for all variants.

gavanderhoorn:
@dniewinski wrote:
> I will move the changes to their own packages [..]

A single package, correct?

And could you please name the package: `ur_e_series_support`? That is in-line with the naming conventions in other ROS-Industrial repositories (ie: `$(mfg)_$(robot_model)_support`).

dniewinski:
I was going to move them to ur_e_description and ur_e_gazebo to keep this repo consistent, even if it isn't following the standard convention.

dniewinski:
And yes, I agree about the meshes.  For whatever reason I was sure that I had checked and the collision meshes were the same as the visual meshes.  Must be getting my packages confused

gavanderhoorn:
@dniewinski wrote:
> I was going to move them to ur_e_description and ur_e_gazebo to keep this repo consistent, even if it isn't following the standard convention.

I don't care too much for consistency (*in this case*) as renaming packages later can be rather tedious, especially if they start to see some use.

If you really feel strongly about it, then we can rename them later.

dniewinski:
Moved everything into ur_u packages

dniewinski:
Would you like me to make new MoveIt! configs for these new arms?  I can either change the robot names to not include the E, which will let the work with the current MoveIt! setups, or I could create all new configs.  What do you prefer?

gavanderhoorn:
I'm a bit on the fence regarding moveit configs: any config we provide in this repository will only be useful for some quick testing, as actual deployments or integrations will require at least a re-run of the MSA.

> I can either change the robot names to not include the E [..]

this would not be an option for me. These are different robots, so they must get unique names.

dniewinski:
Ok, I will make new MoveIt! configs for the new arms.  Agreed, they would just be initial testing and training examples, but they should exist

ur10_e_moveit_config/config/ompl_planning.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - BKPIECEkConfigDefault
+    - KPIECEkConfigDefault
+    - RRTkConfigDefault
+    - RRTConnectkConfigDefault
+    - RRTstarkConfigDefault
+    - TRRTkConfigDefault
+    - PRMkConfigDefault
+    - PRMstarkConfigDefault
+  ##Note: commenting the following line lets moveit chose RRTConnect as default planner rather than LBKPIECE
+  #projection_evaluator: joints(shoulder_pan_joint,shoulder_lift_joint)
```
gavanderhoorn, line 70: Are these copies of the 'regular' moveit configs?

MoveIt supports configuring a default planner now without resorting to this hack, both in the MSA and in `ompl_planning.yaml`. See [fanuc_lrmate200ib3l_moveit_config/config/ompl_planning.yaml](https://github.com/ros-industrial/fanuc/blob/992e4d85d9f026749e90d08e3e509e0ce5e6f5b0/fanuc_lrmate200ib3l_moveit_config/config/ompl_planning.yaml#L56) for an example.

It would be nice if we could do that here as well.

dniewinski: These are just copies of the original ones.  Since they are just base examples for testing and would need to be redone in any actual implementation I only updated minor arm-specific details

gavanderhoorn: That's fine, but let's please set the default planner to `RRTConnect`. We've had quite some difficulties with MoveIt defaulting to `(L)BKPIECE` which does not work very well with the UR arm geometry.

dniewinski:
Default planner updated.  Are you still waiting on me for anything?

gavanderhoorn:
>Are you still waiting on me for anything?

No, just a bit busy and I want to test and check some things myself.

Thanks for the contribution already.

ipa-nhg:
@dniewinski thanks for the contribution, great job! :+1: 

just a comment, could you please add to the metapackage [package.xml](https://github.com/dniewinski/universal_robot/blob/ur_e/universal_robot/package.xml)  the new packages?

PS: I think that in the future we can consider the option of use the robot type names as argument for launch and URDF files and avoid duplicated code. But this PR is consistent with the current status

ipa-nhg (commented):
Test on simulation ur10_e, ur5_e and ur3_e:
- gazebo simulation (visual, collision and controllers) :ok: 
- RVIZ visualization :ok: 
- Moving using rqt trajectory controller :ok: 

The  [package.xml](https://github.com/dniewinski/universal_robot/blob/ur_e/universal_robot/package.xml)  should be updated, see https://github.com/ros-industrial/universal_robot/pull/380#issuecomment-419073885

ipa-nhg (dismissed):

dniewinski:
Sorry, minor update to mesh colours since it hasn't been merged yet

gavanderhoorn:
@dniewinski: I've added 0ff5873 which adds proper collision models.

As noted in https://github.com/ros-industrial/universal_robot/pull/380#issuecomment-416257719, collision meshes should be heavily downsampled versions of the visual ones (but still be a representation of the geometry). There is no need for details in collision models, unless the geometry of the robot requires it (because of holes or arcs fi).

0ff5873 reduces size of the `ur_e_description` pkg by about 8 MB.

The `visual` models are on the large side as well at almost 10 MB per robot variant. That's about 1.5x larger than `ur_description` on average.

Which program / conversion pipeline did you use to create the meshes?

gavanderhoorn:
Note: this PR needs to be squash-merged in order to avoid importing the overly detailed models into the history of the repository.

Repository size is already close to `100MB`. That is quite a lot for a repository with basically a single robot model that it supports (for comparison: `ros-industrial/fanuc` supports 19 different variants and is 23MB, including history).

dniewinski:
I used Solidworks > Blender.  Blender to clean up and split the meshes, set colours, etc.  The collision meshes should be 50-70% lighter than the visual meshes (decimated), but the visuals are still a little heavy.

gavanderhoorn:
> I used Solidworks > Blender. Blender to clean up and split the meshes, set colours, etc. The collision meshes should be 50-70% lighter than the visual meshes (decimated), but the visuals are still a little heavy.

The collison meshes appeared to be decimated versions, but still too detailed. I replaced them with convex hulls. Each of them is now about 80KB.

If needed, we could go for convex decompositions, but we're probably fine like this for now.

AustinDeric:
@dniewinski needs to be rebased due to updates on the target repo. (I opened a PR here https://github.com/dniewinski/universal_robot/pull/1 so all you need to do is approve and that will fix this).

@gavanderhoorn looks like all the outstanding issues have been addressed, after rebase can we merge? We shouldn't have to rebase again, this is a great repo and i think is ready to go.

Tested on actual ur10e, looks good.

dniewinski:
Rebased

AustinDeric:
Awesome, thanks!  @gavanderhoorn can we get a review & merge? If not are there any remaining issues that need to be addressed?

ipa-nhg (approved):

dniewinski:
Hello.  This PR was opened 6 months ago.  Is there any reason it hasn't been merged?  Anything needed from me?

miguelprada:
Sorry to keep this around for so long. Unfortunately, this is now in conflict with the base branch. Would you mind merging `kinetic-devel` or rebasing one last time?

gavanderhoorn:
@miguelprada: please take https://github.com/ros-industrial/universal_robot/pull/380#issuecomment-431405140 into account when merging this.

miguelprada:
Thanks @dniewinski!

PR #391, opened 2018-11-12
MoveGroupExecuteService is Deprecated by MoveIt!

Tianhuanyu:
Replace MoveGroupExecuteService with MoveGroupExecuteTrajectoryAction following the warning.

### DIFF
--- a/ur10_moveit_config/launch/move_group.launch
+++ b/ur10_moveit_config/launch/move_group.launch
@@ -50,7 +50,7 @@
 
     <!-- MoveGroup capabilities to load -->
     <param name="capabilities" value="move_group/MoveGroupCartesianPathService
-				      move_group/MoveGroupExecuteService
+				      move_group/MoveGroupExecuteTrajectoryAction
 				      move_group/MoveGroupKinematicsService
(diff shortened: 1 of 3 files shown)


v4hn (changes requested):
I am not a maintainer of this repo, but one of the MoveIt maintainers, so I guess that counts as a review here :sunglasses: 

Thank you for the patch!

While it is correct that these services are deprecated (and actually removed in melodic),
the correct patch to address this is to get rid of the `capabilities` parameter altogether.
The correct default capabilities are automatically loaded on startup in kinetic+.

Please have a look at [the setup assistant template](https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch#L26) for comparison.

hsd-dev (commented):
Thanks for your PR! Looks good to me as well.

ipa-nhg (approved):

gavanderhoorn (approved):
This is an ok first step. We can address @v4hn's comment in a later PR (the MoveIt configs need to be regenerated anwyay).

PR #392, opened 2018-12-11
Add transmission_hw_interface to UR xacro and expose everywhere

miguelprada:
Adds the capability to select the hardware interface used in the `<transmission>` elements in the URDF model. This allows creating custom simulation setups with different hardware interfaces, e.g. `hardware_interface/VelocityJointInterface`.

One thing I'm not very happy about is repeating the default value in so many places, but I cannot see a way around this to maintain backwards compatibility for users that use the different bits in this repo: e.g. users of the default `ur10_robot.urdf.xacro` vs users that compose other models using the macro itself.

There's also some whitespace changes. Let me know if you prefer I maintain the original lines.

### DIFF
--- a/ur_description/launch/ur10_upload.launch
+++ b/ur_description/launch/ur10_upload.launch
@@ -1,7 +1,8 @@
 <?xml version="1.0"?>
 <launch>
   <arg name="limited" default="false" doc="If true, limits joint range [-PI, PI] on all joints." />
-  
-  <param unless="$(arg limited)" name="robot_description" command="$(find xacro)/xacro --inorder '$(find ur_description)/urdf/ur10_robot.urdf.xacro'" />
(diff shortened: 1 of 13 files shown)


ipa-nhg (approved):
Thanks :+1: 
Tested for the 3 models on simulation, working as expected!

ipa-nhg:
For me looks good, not optimal but you have to define hierarchically a default value for the parameter.  Otherwise, as you already pointed, this will be an issue when you try to compose models.

About the indentation spaces I honestly don't have a clear preference.

gavanderhoorn:
> For me looks good, not optimal but you have to define hierarchically a default value for the parameter. Otherwise, as you already pointed, this will be an issue when you try to compose models.

`xacro` supports default parameters and importing parameters from parent scopes ([wiki/xacro - Default parameters](http://wiki.ros.org/xacro#Default_parameters)). Might be something to see whether that could help here.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @miguelprada.

PR #395, opened 2018-12-19
Add RobotModeDataMsg

felixvd:
See https://github.com/ros-industrial/ur_modern_driver/pull/247

### DIFF
--- a/ur_msgs/CMakeLists.txt
+++ b/ur_msgs/CMakeLists.txt
@@ -15,6 +15,7 @@ add_message_files(
    IOStates.msg
    RobotStateRTMsg.msg
    MasterboardDataMsg.msg
+   RobotModeDataMsg.msg
    ToolDataMsg.msg
 )
 
--- a/ur_msgs/msg/RobotModeDataMsg.msg
+++ b/ur_msgs/msg/RobotModeDataMsg.msg
@@ -1,0 +1,13 @@
+# This data structure contains the RobotModeData structure
+# used by the Universal Robots controller
+#
+# This data structure is send at 10 Hz on TCP port 30002
+
+uint64 timestamp
+bool is_robot_connected
+bool is_real_robot_enabled
+bool is_power_on_robot
+bool is_emergency_stopped
+bool is_protective_stopped
+bool is_program_running
+bool is_program_paused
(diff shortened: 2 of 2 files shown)


miguelprada (commented):
There are some additional fields in the Robot Mode Data, according to the [overview of the client interfaces](https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/remote-control-via-tcpip-16496/).

In particular, I'm interested in the _speed fraction/scaling_ value (name is different depending on version). I'm going to guess that value indicates the position of the slider in PolyScope, but I haven't been able to check this. If that is the case, it could be interesting to add it since application logic may use that to throttle down the commands based on its value.

Any reason for not having included it in the message?

felixvd:
The historical reason is that I merged the fields from [this PR](https://github.com/ros-industrial/ur_modern_driver/pull/181) and didn't want to do unnecessary development that no one might have a need for. I have also had issues with the driver when the slider is not at 100% (e.g. robots would overshoot and not stop at the target position), so adding more fields may open up more construction sites for little benefit. I suspect that in the case you describe, users will either use the URscript commands or send a slower trajectory through ROS.

A practical reason is that the fields that this PR publishes are common to all UR firmware versions, so there is no need to do additional checks.

Either way, new fields can be added afterwards if the need arises.

miguelprada:
I think the slider being not at 100% is a common source of errors, indeed. Anyways, it may make more sense to tackle that problem by adding a check at the driver level and displaying a warning, rather than publishing the value.

I'm fine merging this as it is. @gavanderhoorn, @ipa-nhg?

miguelprada (dismissed):

gavanderhoorn (approved):

gavanderhoorn:
Sorry for dismissing your review @miguelprada, but I felt the additional comment in the `.msg` file would be good to add.

miguelprada:
> Sorry for dismissing your review @miguelprada, but I felt the additional comment in the `.msg` file would be good to add.

:+1: 

Not sure why this cannot be merged as is. I'll merge `kinetic-devel` and squash-merge.

miguelprada:
Thanks @felixvd!

felixvd:
Thanks for the quick action after my late reply!

PR #396, opened 2018-12-19
Joint Names added

fkraemer:
Added joint names config file as in other ros_industrial robot support packages.


ipa-nhg:
@fkraemer why are you adding these files here? where are you loading this yamls files (I mean launch file)? 

NOTE: The configuration of the controllers for the URs is already located in the `ur_modern_driver` https://github.com/ros-industrial/ur_modern_driver/tree/master/config

fkraemer:
> @fkraemer why are you adding these files here? where are you loading this yamls files (I mean launch file)?
> 
> NOTE: The configuration of the controllers for the URs is already located in the `ur_modern_driver` https://github.com/ros-industrial/ur_modern_driver/tree/master/config

This is required by moveIt! (needs specific ros parameters loaded from the added files if joints have non standard names). See: http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot -> joint_names.yaml

The exact location in the file tree can be discussed of course.

gavanderhoorn:
> > @fkraemer why are you adding these files here? where are you loading this yamls files (I mean launch file)?
> > NOTE: The configuration of the controllers for the URs is already located in the `ur_modern_driver` https://github.com/ros-industrial/ur_modern_driver/tree/master/config
> 
> This is required by moveIt! (needs specific ros parameters loaded from the added files if joints have non standard names). See: http://wiki.ros.org/Industrial/Tutorials/Create_a_MoveIt_Pkg_for_an_Industrial_Robot -> joint_names.yaml

Just a bit of clarification: MoveIt requires names of `joint`s for the configuration of its controllers (simple moveit controller fi), but it doesn't use the `controller_joint_names` variable for that.

I don't believe the MoveIt configurations in this repository are setup to use `controller_joint_names`, so while it could potentially be a good idea from a consistency perspective (other robot support packages do have this variable), functionally it would not seem to be needed.

gavanderhoorn:
First: thanks for making the effort to contribute and submitting the PR @fkraemer.

However: as none of the packages in this repository use `controller_joint_names`, nor `ur_modern_driver`, I don't believe we should merge this.

In addition to it being unused in this repository, it would also introduce additional maintenance, as use of `ros_control` would mean we'd be maintaining two parameters with identical contents.

gavanderhoorn:
I'd like consensus, so @ipa-nhg, @miguelprada: opinions?

miguelprada:
While any specific need for this parameter doesn't show up, I agree with you, @gavanderhoorn.

PR #398, opened 2019-01-15
Updated xacro namespace.

darkdragon-001:
Use new namespace to avoid _inconsistent namespace redifinitions for xmlns:xacro_ warnings.

### DIFF
--- a/ur_description/urdf/common.gazebo.xacro
+++ b/ur_description/urdf/common.gazebo.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
   <gazebo>
     <plugin name="ros_control" filename="libgazebo_ros_control.so">
--- a/ur_description/urdf/ur.gazebo.xacro
+++ b/ur_description/urdf/ur.gazebo.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
   <xacro:macro name="ur_arm_gazebo" params="prefix">
 
--- a/ur_description/urdf/ur.transmission.xacro
+++ b/ur_description/urdf/ur.transmission.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
   <xacro:macro name="ur_arm_transmission" params="prefix">
 
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
 <!--
   Author: Kelsey Hawkins
--- a/ur_description/urdf/ur10_joint_limited_robot.urdf.xacro
+++ b/ur_description/urdf/ur10_joint_limited_robot.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro"
+<robot xmlns:xacro="http://wiki.ros.org/xacro"
        name="ur10" >
 
   <!-- common stuff -->
--- a/ur_description/urdf/ur10_robot.urdf.xacro
+++ b/ur_description/urdf/ur10_robot.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro"
+<robot xmlns:xacro="http://wiki.ros.org/xacro"
        name="ur10" >
 
   <!-- common stuff -->
--- a/ur_description/urdf/ur3.urdf.xacro
+++ b/ur_description/urdf/ur3.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
 <!--
   Author: Felix Messmer
--- a/ur_description/urdf/ur3_joint_limited_robot.urdf.xacro
+++ b/ur_description/urdf/ur3_joint_limited_robot.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro"
+<robot xmlns:xacro="http://wiki.ros.org/xacro"
        name="ur3" >
 
   <!-- common stuff -->
--- a/ur_description/urdf/ur3_robot.urdf.xacro
+++ b/ur_description/urdf/ur3_robot.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro"
+<robot xmlns:xacro="http://wiki.ros.org/xacro"
        name="ur3" >
 
   <!-- common stuff -->
--- a/ur_description/urdf/ur5.urdf.xacro
+++ b/ur_description/urdf/ur5.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<robot xmlns:xacro="http://wiki.ros.org/xacro">
 
   <xacro:include filename="$(find ur_description)/urdf/ur.transmission.xacro" />
   <xacro:include filename="$(find ur_description)/urdf/ur.gazebo.xacro" />
--- a/ur_description/urdf/ur5_joint_limited_robot.urdf.xacro
+++ b/ur_description/urdf/ur5_joint_limited_robot.urdf.xacro
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<robot xmlns:xacro="http://ros.org/wiki/xacro"
+<robot xmlns:xacro="http://wiki.ros.org/xacro"
        name="ur5" >
 
   <!-- common stuff -->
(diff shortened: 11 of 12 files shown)


darkdragon-001:
Can anyone please review?

ipa-nhg (approved):
Thanks! :+1:

v4hn:
Hi there,

as @lianghongzhuo just pointed out to me, there seem to be at least three different URLs in existing xacro files to define the namespace, all in use in active development and even likely to be combined by different groups.

1. [`http://ros.org/wiki/xacro`](https://github.com/ros-industrial/robotiq/blob/kinetic-devel/robotiq_2f_140_gripper_visualization/urdf/robotiq_arg2f_140_model_macro.xacro#L2)
2. [`http://wiki.ros.org/xacro`](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_description/urdf/common.gazebo.xacro#L2)
3. [`http://www.ros.org/wiki/xacro`](https://github.com/shadow-robot/sr_common/blob/kinetic-devel/sr_description/robots/arm_and_hand_motor.urdf.xacro#L1)

One of the downsides of http-redirects I guess. :)

While this pull-request uses `http://wiki.ros.org/xacro` , we updated the PR2 repositories recently to `http://www.ros.org/wiki/xacro` [here](https://github.com/PR2/pr2_common/pull/270/files), because this is the namespace given in the xacro tutorial.

Could we maybe agree on a single URL [to avoid redundant pull-requests](https://github.com/TAMS-Group/tams_ur5/pull/6) to streamline this?
Alternatively we should really consider patching xacro to avoid the warning altogether.

@gavanderhoorn

darkdragon-001:
I used `http://wiki.ros.org/xacro` because this is where all URLs redirect to, so it seems to be the preferred one.

mintar:
Yeah, we're in a bit of a pickle here. Here's my take on the three options:

### 1. http://ros.org/wiki/xacro

* used in 85% of xacro files on my system (from /opt/ros/kinetic)

### 2. http://www.ros.org/wiki/xacro

Recommended by both tutorials:

* http://wiki.ros.org/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File
* http://gazebosim.org/tutorials/?tut=ros_urdf

It seems [somebody just put it into the ros.org tutorial in 2012](http://wiki.ros.org/action/diff/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File?action=diff&rev1=5&rev2=6), even though it wasn't standard usage at the time, and the gazebosim.org tutorial just copied it from there.

### 3. http://wiki.ros.org/xacro

* used by no-one

----------------------------------------------------------------------

So that rules out (3), but which one to pick between (1) and (2)? The problem is that everybody should be consistent, otherwise there will be warnings about inconsistent namespaces if one xacro file includes another one with a different namespace. My impression is that (1) was used historically and *should* be the default, but the tutorials recommended (2) for the past 7 years, and now some people are starting to switch over to (2), which adds to the confusion. So I'm going with (1) for now, since that is what most packages use currently, and what's used in this repo as well. Therefore I would suggest to keep the namespaces as they are.

gavanderhoorn:
@mintar wrote:
> ### 3. http://wiki.ros.org/xacro
> 
>     * used by no-one

That's not actually true: the `fanuc` and `fanuc_experimental` pkgs use that one -- and there are some others (which are not robot support pkgs).

And personally I feel that one makes the most sense as well: I know it doesn't *really* matter, but `wiki.ros.org/xacro` is the official URL.

mintar:
Oh damn. So there are actually users of all 3 versions. Personally, I don't really care, but those "inconsistent namespace" warnings will only go away if everyone globally uses the same namespace...

Since it doesn't really matter anyway, maybe we can just remove the pesky warning from xacro and not care about the namespaces?

https://github.com/ros/xacro/blob/a19a4f38ca0957391933725564d2a5034de00ea1/src/xacro/__init__.py#L399

mintar:
Anyway, I would suggest just leaving all namespaces the way they are, at least until we can come up with a clear answer as to which version is the only correct one. Otherwise, as soon as one package changes the namespaces, it will fix the warnings for some users at the cost of triggering the warnings for other users.

gavanderhoorn:
I know it's only going to go away once we get everything in sync, but that's probably never going to happen: there are just too many packages with either of the three variants.

If the URL is supposed to make any sense, I'd vote for 3 as that is currently the canonical URL.

XML ns URIs are only used to discriminate between ns (it's just a bit of text that is supposed to be unique to avoid clashes), but at least option 3 has some sort of rational argument -- other than historical usage.

> Otherwise, as soon as one package changes the namespaces, it will fix the warnings for some users at the cost of triggering the warnings for other users.

yes, and that has happened already many times (for me and others).

mintar:
> XML ns URIs are only used to discriminate between ns (it's just a bit of text that is supposed to be unique to avoid clashes), but at least option 3 has some sort of rational argument -- other than historical usage.

Well, you could say the same for the other two options:

(1) seems to be the most commonly used variant (by my guesstimate), which is a strong argument if the goal is to get everything consistent
(2) is the variant recommended by the official tutorials

gavanderhoorn:
True, but a counter could be: the tutorials should have been updated when the official URL changed and all those other pkgs should be updated as well.

The last is not very realistic -- I admit. The tutorial could be updated so at least new files will use the "correct" URL.

Seeing as it's just a bit of text anyway, I would actually propose we revert this PR and keep `http://ros.org/wiki/xacro` as the `ns`.

But I'd like the input of some the other maintainers: @ipa-nhg @miguelprada?

---

Edit: one use of `xmlns` that we're ignoring here is "self documentation": the URL used to define the `ns` can also point to the place where more information may be found about the elements that are part of the namespace. In that case using the canonical URL would seem to make sense again, but we get back to the infeasibility of updating pkgs that are using stale ones.

mintar:
The strongest argument for me is that if you're going to introducing a breaking change in existing behavior, there better be a good argument that you're actually fixing incorrect behavior instead of just personal preference. I don't see a compelling argument that either of the three options is the authoritative one.

>  I would actually propose we revert this PR

Damnit, I totally missed that this PR has already been merged.

Seriously: Shouldn't we just remove the warning in xacro, as I proposed above? I know it seems like a cop-out, but I believe this is the best option (since we're never going to get the namespaces to be consistent). Currently, the warning mainly confuses new users, who then go submit PRs to random repos changing the namespace (I've seen plenty of those). Or at least the warning should clearly state which namespace is the authoritative one, so that all PRs go in the same direction.

@rhaschke seems to be the maintainer of xacro these days. Any opinions?

miguelprada:
Considering this PR has already been merged and points to the _canonical_ URL, I'd also vote for keeping it pointing to `http://wiki.ros.org/xacro`. The "self documentation" argument also favors this, even if users will eventually get redirected to the right URL. Not a strong opinion, anyways.

> Seriously: Shouldn't we just remove the warning in xacro, as I proposed above? I know it seems like a cop-out, but I believe this is the best option (since we're never going to get the namespaces to be consistent). Currently, the warning mainly confuses new users, who then go submit PRs to random repos changing the namespace (I've seen plenty of those). Or at least the warning should clearly state which namespace is the authoritative one, so that all PRs go in the same direction.

I'm not familiar with the `xacro` codebase. Is this even possible? Or is the warning raised by the underlying XML handling code when finding a ns redefined? A warning message pointing to the "correct" `xacro` namespace would definitely help in eventually having all repos use the same namespace. Otherwise, I would vote for removing the warning altogether as well.

mintar:
> I'm not familiar with the xacro codebase. Is this even possible? Or is the warning raised by the underlying XML handling code when finding a ns redefined?

It's printed by xacro itself, in this line:

https://github.com/ros/xacro/blob/a19a4f38ca0957391933725564d2a5034de00ea1/src/xacro/__init__.py#L399

rhaschke:
I don't have a specific preference for any of the listed variants. Even, in the xacro unit tests we don't use a unique url :frowning_face:
If we agree on one of those, adding a hint in the xacro warning shouldn't be a big issue, although the warning more generally targets any namespace conflicts. For this reason, I cannot agree to remove it completely. Actually, any xacro namespace declaration should be removed from the processed file as the final final isn't xacro anymore. There is an open issue for this, but I didn't have time to tackle this yet.

gavanderhoorn:
Technically, a `xacro:macro` with `xmlns="some_value"` and a `xacro:macro` with `xmlns="other_value"` are not to be considered the same `xacro:macro`. The fact that we (as users) know the same is meant is irrelevant. So I understand why the warning is there (it should perhaps even be an error).

As to the issue at hand: if we go by the nrs, *probably* we should use `http://ros.org/wiki/xacro` (it's just "a piece of text" anyway, so there's no need for it to correspond to any real URL or a valid one).

> Actually, any xacro namespace declaration should be removed from the processed file as the final final isn't xacro anymore. There is an open issue for this, but I didn't have time to tackle this yet.

This one: https://github.com/ros/xacro/issues/144 right?

rhaschke:
> Technically, a xacro:macro with xmlns="some_value" and a xacro:macro with xmlns="other_value" are not to be considered the same xacro:macro.

Exactly.

> This one: [ros/xacro#144](https://github.com/ros/xacro/issues/144) right?

Yep. Thanks for adding the link.

v4hn:
> Technically, a xacro:macro with xmlns="some_value" and a xacro:macro with xmlns="other_value" are not to be considered the same xacro:macro.

Yes. I very much agree that this should be respected for the general case.

It would be possible though to add these three concrete strings as a "301 redirect" equivalence class in the semantic interpretation of xacro, treating all three equal.
Would that be an acceptable solution for upstream?

rhaschke:
I think, fixing https://github.com/ros/xacro/issues/144, will also resolve the xacro namespace issues as then those namespace definition will not show up in the final document and thus cannot produces clashes anymore. I would prefer this solution, instead of adding an equivalence classes of xacro namespaces. If anybody volunteers to contribute a xacro PR for this, you are very welcome.
I will not have the time to tackle this within the next few weeks at least.
It's not too complicated, just remove all namespace definitions within the equivalence class (or even all mentioning xacro?) from the final document.

gavanderhoorn:
@rhaschke wrote:
> I think, fixing [ros/xacro#144](https://github.com/ros/xacro/issues/144), will also resolve the xacro namespace issues as then those namespace definition will not show up in the final document and thus cannot produces clashes anymore.

I'm slightly confused: isn't `xacro` printing these warnings?

rhaschke:
Yes, xacro is printing these warnings for all namespace definitions causing clashes (in value) after having collected all the definitions from the individual files.
Addressing https://github.com/ros/xacro/issues/144 will filter out all namespace definitions for `xmlns:xacro` (regardless of their value) and thus not cause any clashes anymore.

rhaschke:
@gavanderhoorn I found some time to address https://github.com/ros/xacro/issues/144 in https://github.com/ros/xacro/pull/207. Could you please have a look and verify that this also resolve this issue?

miguelprada:
I can confirm that with the changes in ros/xacro#207 the warnings indeed go away.

PR #401, opened 2019-02-07
rename metapackage name, from universal_robot to universal_robots

ipa-nhg:


gavanderhoorn:
Thanks for the PR @ipa-nhg.

May I suggest we first *add* a new metapackage `universal_robots` and update the `universal_robot` metapackage to depend on that instead of the actual packages? That would make sure to not break installation instructions not under our control and packages of other users that depend on the metapackage (they shouldn't, but its all too common).

After sufficient time we can then remove the `universal_robot` metapackage.

ipa-nhg:
@gavanderhoorn like this https://github.com/ros-industrial/universal_robot/pull/409 ?

See https://github.com/ros-industrial/universal_robot/blob/fa35caa79ffcddec6b1981d540cbbe565a56e5bb/universal_robots/package.xml#L22

gavanderhoorn:
I've added some comments to #409.

gavanderhoorn:
Shall we close this PR in favour of #409?

ipa-nhg:
close, superseded by #409

Issue #403, opened 2019-02-17
moveit kdl_kinematics_plugin includes cause build failure


jwhendy:
I think I'm having an issue similar to #274 . 

I'm building melodic on arch linux with cmake 3.13 and boost 1.69. I have to remove these [two lines](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_kinematics/include/ur_kinematics/ur_moveit_plugin.h#L97) in order to build:
```
#include <moveit/kdl_kinematics_plugin/chainiksolver_pos_nr_jl_mimic.hpp>
#include <moveit/kdl_kinematics_plugin/chainiksolver_vel_pinv_mimic.hpp>
```

Errors:
```
Errors     << ur_kinematics:make /home/jwhendy/catkin_ws/logs/ur_kinematics/build.make.008.log                                                                                           
In file included from /home/jwhendy/catkin_ws/src/universal_robot/ur_kinematics/src/ur_moveit_plugin.cpp:75:
/opt/ros/melodic/include/class_loader/class_loader.h:36:2: warning: #warning Including header <class_loader/class_loader.h> is deprecated, include <class_loader/class_loader.hpp> instead. [-Wcpp]
 #warning Including header <class_loader/class_loader.h> is deprecated, \
  ^~~~~~~
In file included from /home/jwhendy/catkin_ws/src/universal_robot/ur_kinematics/src/ur_moveit_plugin.cpp:75:
/opt/ros/melodic/include/class_loader/class_loader.h:36:2: warning: #warning Including header <class_loader/class_loader.h> is deprecated, include <class_loader/class_loader.hpp> instead. [-Wcpp]
 #warning Including header <class_loader/class_loader.h> is deprecated, \
  ^~~~~~~
In file included from /home/jwhendy/catkin_ws/src/universal_robot/ur_kinematics/src/ur_moveit_plugin.cpp:75:
/opt/ros/melodic/include/class_loader/class_loader.h:36:2: warning: #warning Including header <class_loader/class_loader.h> is deprecated, include <class_loader/class_loader.hpp> instead. [-Wcpp]
 #warning Including header <class_loader/class_loader.h> is deprecated, \
  ^~~~~~~
In file included from /home/jwhendy/catkin_ws/src/universal_robot/ur_kinematics/src/ur_moveit_plugin.cpp:88:
[...]

PR #404, opened 2019-02-17
removed non-existent moveit KDL libraries from ur_kinematics includes

jwhendy:
Per [this moveit issue](https://github.com/ros-planning/moveit/issues/1351) it seems that #403 should be handled by removing these includes. This is a PR removes the non-existend KDL headers from the ur_kinematics includes so that this builds on melodic.


miguelprada (approved):
Even if this fix is only required for melodic, since those [aren't supposed to be public headers](https://github.com/ros-planning/moveit/issues/1351#issuecomment-464454344) and the package builds without including them, I'd say it's ok to merge.

gavanderhoorn:
Feel free to merge @miguelprada.

gavanderhoorn (approved):

miguelprada:
Thanks @jwhendy!

iameijaz:
can this work for Noetic as well?

PR #406, opened 2019-03-04
Require kinetic checks to pass and add optional melodic checks

miguelprada:
Checks for `kinetic` were still allowed to fail. This makes them required and adds new _optional_ `melodic` checks.


ipa-nhg (approved):
Thanks! :+1:

gavanderhoorn (approved):

PR #409, opened 2019-03-07
add new metapackage, first step to rename the repo

ipa-nhg:


gavanderhoorn:
I've added some commits.

This PR now introduces a `universal_robots` metapackage that depends on all the other UR pkgs, and makes `universal_robot` depend on `universal_robots`.

People that use old installation instructions install `universal_robot`, which will install `universal_robots` and all the other packages.

Updated instructions should use the new name, which will also install all the UR packages.

At some point in the future we can remove `universal_robot`.

ipa-nhg:
@gavanderhoorn ah ok, I understand now the point! Looks good to me! Thanks! :+1:

gavanderhoorn (approved):

miguelprada (approved):
Looks good to me.

PR #415, opened 2019-04-02
msgs: clarify use of fields in SetIO svc

gavanderhoorn:
As per subject.

### DIFF
--- a/ur_msgs/srv/SetIO.srv
+++ b/ur_msgs/srv/SetIO.srv
@@ -1,7 +1,40 @@
+# Service allows setting digital (DIO) and analog (AIO) IO, as well as flags
+# and configuring tool voltage.
+#
+# This service has three request fields (see below).
+#
+# When setting DIO or AIO pins to new values:
+#
+#  fun    The function to perform
+#  pin    Numeric ID of the pin to set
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Triggered by [UR10 tool flange 24V power control](https://answers.ros.org/question/320110) on ROS Answers.

miguelprada (approved):
Nice, thanks @gavanderhoorn!

miguelprada:
Will merge when the tests pass.

It makes me think whether `ur_modern_driver` should use `std::lround` instead of `static_cast<uint8_t>` to [transform tool voltages](https://github.com/ros-industrial/ur_modern_driver/blob/7fae9f236e9557eb09e5e6554abe54da9de8f24a/include/ur_modern_driver/ros/io_service.h#L53) just to be safer.

gavanderhoorn:
> It makes me think whether `ur_modern_driver` should use `std::lround` instead of `static_cast<uint8_t>` to [transform tool voltages](https://github.com/ros-industrial/ur_modern_driver/blob/7fae9f236e9557eb09e5e6554abe54da9de8f24a/include/ur_modern_driver/ros/io_service.h#L53) just to be safer.

It should at the very least probably check whether `state in [0, 12, 24]`.

But yes, that would make sense.

Can you log an issue over at `ros-industrial/ur_modern_driver`?

PR #416, opened 2019-04-04
Correct width and type of 'digital_*_bits' fields in MasterBoardData

gavanderhoorn:
As per subject.

It might be that the upper word of these fields was 'reserved' in older protocol descriptions, but in all versions in which these fields are present they are now 4 byte integers.


miguelprada:
Looks good to me :+1:

miguelprada (approved):
Looks good to me :+1:

PR #417, opened 2019-04-04
travis: remove Indigo configurations.

gavanderhoorn:
As per subject.

It will most likely stay compatible with Indigo for a long time, but we're not going to spend effort trying to maintain that.


miguelprada (approved):
Makes sense.

PR #422, opened 2019-04-29
Load the JointGroupPositionController so jog commands can be sent

AndyZe:
I would like this so I can finish a jogging tutorial based on the UR3 in Gazebo

### DIFF
--- a/ur_e_gazebo/controller/arm_controller_ur10e.yaml
+++ b/ur_e_gazebo/controller/arm_controller_ur10e.yaml
@@ -19,3 +19,13 @@ arm_controller:
   stop_trajectory_duration: 0.5
   state_publish_rate:  25
   action_monitor_rate: 10
+joint_group_position_controller:
+  type: position_controllers/JointGroupPositionController
+  joints:
+     - shoulder_pan_joint
+     - shoulder_lift_joint
+     - elbow_joint
+     - wrist_1_joint
+     - wrist_2_joint
+     - wrist_3_joint
+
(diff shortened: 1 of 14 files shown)


AndyZe:
I guess it would make sense to add the same for UR5/UR10, too, right?

gavanderhoorn:
> I guess it would make sense to add the same for UR5/UR10, too, right?

That does make sense, but then it would also make sense to load the controllers. The current PR only does that for the UR3.

mlautman:
looks good to me

AndyZe:
@gavanderhoorn @ipa-nhg @miguelprada Not to pressure/stress you, but this is the last call before I move forward with a different Gazebo robot

miguelprada (changes requested):
Sorry for the delayed response.

The addition of the new controller demands `position_controllers` to be declared as a `run_depend` in the package manifest.

Also, for completeness sake, it'd be good to replicate the changes in `ur_e_gazebo` (don't forget the dependency there as well).

miguelprada (approved):

miguelprada:
Looks good to me. @gavanderhoorn, @ipa-nhg could you give it a second look and merge if you're ok with it?

mlautman (approved):

gavanderhoorn (approved):
Might be nice to sync the `ros_control` configs for the real robot and the Gazebo ones, but this is OK for now.

gavanderhoorn:
@miguelprada: feel free to merge.

gavanderhoorn:
And please squash merge. The fixups don't need to become part of the history.

miguelprada:
Thanks for iterating and for your patience @AndyZe!

PR #426, opened 2019-05-29
corrected dimensions and positions of inertias

fmauch:
I'm by far not an expert in working with gazebo or inertias, but it seemed wrong to me:

- The upper arm inertia of the ur10 is not centered in the visual arm segment
- CoM in the wrist links don't sit inside the correct links. E.g. wrist1 has its CoM inside the end of the forearm for all robots.
- Because of the second point the inertia's geometry of wrist3 is matching wrist2 instead of the actual moving part of wrist3.
- Wrist dimensions of ur5 were completely off.
- On the ur5e the arm inertias weren't centered in the visuals.

### DIFF
--- a/ur_description/urdf/ur10.urdf.xacro
+++ b/ur_description/urdf/ur10.urdf.xacro
@@ -60,6 +60,7 @@
     <!-- Arbitrary offsets for shoulder/elbow joints -->
     <xacro:property name="shoulder_offset" value="0.220941" />  <!-- measured from model -->
     <xacro:property name="elbow_offset" value="-0.1719" /> <!-- measured from model -->
+    <xacro:property name="upper_arm_inertia_offset" value="0.045" />  <!-- measured from model -->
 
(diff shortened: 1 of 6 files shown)


miguelprada (approved):
Not sure how meaningful these changes are, since the UR simulation only exposes a _position_ interface which should mostly disregard the inertial parameters.

This anyways seems somewhat better than before, so I'd vote for a merge so we can move forward to #414, which is actually quite important.

gavanderhoorn:
@miguelprada wrote:
> Not sure how meaningful these changes are, since the UR simulation only exposes a _position_ interface which should mostly disregard the inertial parameters.

true, but moving to effort controllers is on (my personal) todo/wish-list, so this would be important.

gavanderhoorn (approved):

gavanderhoorn:
Thanks for the PR @fmauch :+1:

PR #435, opened 2019-07-02
Add description view files

fmauch:
This MR adds launch files for viewing the individual descriptions with zero overhead.

### DIFF
--- a/ur_description/launch/view_ur10.launch
+++ b/ur_description/launch/view_ur10.launch
@@ -1,8 +1,10 @@
 <?xml version="1.0"?>
 <launch>
+  <include file="$(find ur_description)/launch/ur10_upload.launch"/>
+
   <node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" >
     <param name="use_gui" value="true"/>
   </node>
   <node name="robot_state_publisher" pkg="robot_state_publisher" type="state_publisher" />
-  <node name="rviz" pkg="rviz" type="rviz" args="-d $(find urdf_tutorial)/urdf.rviz" />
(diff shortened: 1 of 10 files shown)


miguelprada (dismissed):
Thanks @fmauch!

ur_description/launch/view_ur10.launch
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<launch>
+  <include file="$(find ur_description)/launch/ur10_upload.launch"/>
+
+  <node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" >
+    <param name="use_gui" value="true"/>
+  </node>
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="state_publisher" />
+  <node name="rviz" pkg="rviz" type="rviz" args="-d $(find ur_e_description)/etc/view_robot.rviz" required="true" />
```
gavanderhoorn: this is a `.launch` file in `ur_description`, but it uses the RViz config from `ur_e_description`, even though `ur_description` also has an RViz config now.

Was that intended?

gavanderhoorn: Additionally: could we place the RViz config in a `cfg` sub dir? Instead of the `etc`?

gavanderhoorn (changes requested):
If you could address the two small comments this could be merged.

Nice contribution :+1:

ur_description/launch/view_ur10.launch
fmauch: I changed the files accordingly

gavanderhoorn (approved):
Thanks again for the PR :+1:

gavanderhoorn (approved):
Thanks again :+1: 

looks like travis finally caught up.

gavanderhoorn:
Thanks @fmauch.

:+1: :medal_sports:

Issue #442, opened 2019-07-04
`ur_kinematics` issues and future work?


Yuki-cpp:
We were looking at `ur_kinematics` and noticed that it seems to have received little attention over the last year despite a number of issues. A project has been created that consolidates all of these issues, but is there a roadmap to address them? If not, we will modify `ur_kinematics` to fit our own purposes, including a fairly large refactoring that would probably break existing API but would make the package more functional. Would you be interested in this sort of work?

gavanderhoorn:
Yes, there would certainly be interest.

However, `ur_kinematics` is first and foremost a MoveIt IK plugin for the UR MoveIt packages. That functionality must be preserved in some way (through a shim or some other way).

Second: the currently outstanding PRs seem to resolve quite a few issues and I'd like those to be integrated instead of ignored.

re: planning: properly dealing with IK-related issues is non-trivial and they potentially have a large impact if handled incorrectly. There simply hasn't been someone taking this on. If you'd be willing to do that, I'd fully support it.

haudren:
I'm piggybacking on this issue to give a little update:
- We started some work on this, we'll keep you updated as soon as we have something worth showing
- I noticed that #358 would bring a lot of improvements, any idea what's blocking here? We would like to base our improvements on it.

Thanks,

gavanderhoorn:
> We started some work on this, we'll keep you updated as soon as we have something worth showing

may I suggest to provide updates sooner and more often?

It would be very unfortunate if you spend a lot of effort only to not pass a (PR) review later.

> I noticed that #358 would bring a lot of improvements, any idea what's blocking here? We would like to base our improvements on it.

As I wrote in my first response:

> re: planning: properly dealing with IK-related issues is non-trivial and they potentially have a large impact if handled incorrectly. **There simply hasn't been someone taking this on.**

It's a matter of resources.

gavanderhoorn:
@Yuki-cpp @haudren: have you made any progress?

gavanderhoorn:
Ping @Yuki-cpp and @haudren.

PR #453, opened 2019-09-04
Update ur_kinematics

Yuki-cpp:
## Summary ##
`ur_kinematics` can only support one UR robot type without being recompiled and uses old style C++ code. Moreover the python bindings are not functional. It also doesn't support the e-series robots. In this PR, we aim at modernizing this package, fixing those 3 issues.

## Details ##
Here is what was done:
- Updated the UR robot parameter handling system to support robot selection at run-time instead of compile time. Added parameters for the e-series robots
- Added pybind11 python bindings
- Added a solution exploration function to expand the 8 original solutions into the full set  of solutions

## Warning ##
Since robot parameters are handled at run-time instead of compile time, a small over-head occurs reducing the time performances by a few percents. However, with C++20 we could remove said over head with the addition of class types in non-type template parameters

Related issue : #442


gavanderhoorn:
First: thank you for the PR and the work, this looks like a good improvement.

Three initial comments:

 - as mentioned in https://github.com/ros-industrial/universal_robot/issues/442#issuecomment-509122949, some discussion beforehand would have been good here. As an example, we're targeting Kinetic as well as Melodic with this repository. For Kinetic, the highest C++ standard we can require is 11, not 14. It may work on Kinetic (with GCC fi), but that is not the point. Going to C++20 would not be possible, even for Melodic.
   This is not necessarily a problem, but it would have been nice to avoid even having to discuss this, which was the idea when I wrote https://github.com/ros-industrial/universal_robot/issues/442#issuecomment-509122949.
 - this seems to overlap and/or replace #305 and #316 (and by extension #358 and #355). Can you comment on this?
 - we'd like to avoid the use of submodules here. I'm not sure how we'll get `pybind11` in, but if possible, submodules would not be the way.

gavanderhoorn:
Apparently there is a Catkin-wrapped version of `pybind11`: [ipab-slmc/pybind11_catkin](https://github.com/ipab-slmc/pybind11_catkin). It's also been released into both Kinetic and Melodic.

gavanderhoorn:
Starting with Bionic we could depend on https://github.com/ros/rosdistro/pull/22261.

jbeck28:
I'm having issues using the kinematics plugin for a UR5e. Specifically, it seems the included `kinematics.yaml` is actually using KDL rather than the UR5e Kinematics plugin. When I modified this through setup assistant to use the UR5e kinematics plugin, I found that upon moving the interactive marker in RViz, the end effector becomes strangely oriented, and I can never find non-approximate IK solutions. 

Is this something anyone else has encountered?

PR #454, opened 2019-09-12
Added a service to set the speed slider position

fmauch:
In the new driver we offer a service to set the speed slider position from ROS side. I think, this would be a good place to keep that service.


ur_msgs/srv/SetSpeedSlider.srv
```diff
@@ -0,0 +1,3 @@
+float64 data
```
gavanderhoorn: Could we give this a better name? `percentage` maybe? Or perhaps `ratio`?

Does UR have a name for this? If so we should reuse it.

ur_msgs/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ add_message_files(
 add_service_files(
    FILES
    SetPayload.srv
+   SetSpeedSlider.srv
```
gavanderhoorn: Is the official name "speed slider"? I seem to remember a different name.

As "speed slider" is not a generally understood term, it would be good to use either the official nomenclature, or a descriptive name.

ur_msgs/srv/SetSpeedSlider.srv
fmauch: I changed it to use the term from the RTDE interface: https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/real-time-data-exchange-rtde-guide-22229/

ur_msgs/CMakeLists.txt
fmauch: It's called "speed_slider_fraction" in the interface, so I stuck to that name. We could also call it `SetSpeedSliderPosition` to be more precices, but that's a rather long name.

gavanderhoorn: I would be fine with `SetSpeedSliderPosition`.

But does it really set the position, or does it set the 'speed fraction' for the internal control system (and is the speed slider only a UI thing for humans)?

fmauch: It sets the fraction of the speed being used to command (in a user perspective) the robot, which is equivalent to the slider position in the UI on the teach panel.

The speed used to command the robot is then a combination from this fraction and the speed scaling active.

If for example, the speed slider (speed_slider_fraction) is set to 0.5 and speed scaling due to safety limitations reduces speed to 0.5, as well, the speed used for control will be 0.25.

gavanderhoorn: I know how it works internally :)

What I was asking about is: *if* it only sets the slider position, which then updates the internal fraction, the current name is OK. If however it actually sets the fraction itself, and the slider is updated by some system process to inform the human, the current name would be slightly incorrect.

It's minor though.

fmauch: That's why we named it `SetSpeedSlider` only in first place :)

So, `SetSpeedSliderFraction` would probably be technically correct. Should we go for that?

gavanderhoorn: > So, `SetSpeedSliderFraction` would probably be technically correct. Should we go for that?

If it sets the fraction, then yes.

Otherwise the current name would be OK.

fmauch: Thanks @gavanderhoorn I think, this discussion is then resolved.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @fmauch for the contribution :+1:

Issue #448, opened 2019-07-13
Breaking changes in Melodic release


gavanderhoorn:
This is a meta-issue / discussion issue for the first Melodic release, which is planned to include a number of breaking changes.

Summarising for now:

 - `universal_robot` (metapackage) (removed in #499):
   - will be removed (`universal_robots` replaces it, `universal_robot` depends on `universal_robots` right now)
 - `ur_driver` (#440, removed in #498)
   - will be removed: `ur_modern_driver` will become the standard driver for CB1 and CB2 controllers, `ur_robot_driver` will be the driver for CB3 and e-series controllers
 - `ur_bringup` (#429, #440, removed in #498)
   - will be removed (instead: drivers should provide the necessary `.launch` files)
 - MoveIt configuration packages:
   - updated to include latest MoveIt functionality (#430)
   - files and layout will follow ROS-I standards (`moveit_planning_execution.launch`, etc) (#431)
 - `ur_description`:
   - support for calibration data imported from robot controller (#414, #357)
   - "joint limited" variants will be removed
   - migration to ROS-I standard structure with top-level and `_macro.xacro` and standard `.launch` files (started with #435, completed in #497)
   - heavy use of `xacro` yaml support (#371)
   - single pair of top-level and `_macro.xacro` which will be parameterised with `.yaml` settings files: one for each UR model (both e-series and regular robots) (#371 and #497)
   - removal of the `world` link (#284 and https://github.com/ros-industrial/universal_robot/pull/371#issuecomment-510568694)
   - removal of all Gazebo-specific infrastructure and settings (will be migrated to `ur_gazebo`)
 - `ur_e_description`:
   - merged with `ur_description`
 - `ur_gazebo` (mostly done in #518):
   - will use similar `.yaml` based approach as `ur_description`
   - will provide a `_macro.xacro` that adds Gazebo-specific elements and plugins (similar to (but not necessarily identical to) [ros-industrial/abb_experimental/abb_irb120_gazebo](https://github.com/ros-industrial/abb_experimental/tree/35af6b1a359777052292dfae834535d79bc4d725/abb_irb120_gazebo))
   - will provide `.launch` files to load `robot_description` and other parameters (`gazebo_ros_control` configuration fi)
   - migration to `ros_control` effort controllers for all simulated robots (#521)
 - `ur_e_gazebo`:
   - will be merged with `ur_gazebo`
 - `ur_msgs`:
[...]

PR #457, opened 2019-09-29
Modernize python 2 codes

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/ur_driver/src/ur_driver/deserialize.py
+++ b/ur_driver/src/ur_driver/deserialize.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import struct
 
 class PackageType(object):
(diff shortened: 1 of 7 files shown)


gavanderhoorn (approved):
`ur_driver` is deprecated and will be removed (#448), but we can merge this as `kinetic-devel` will be kept around for those who want to still use it.

gavanderhoorn:
Seems the new Travis integration is not picking up on this one.

Merging as the old `travis-ci/pr` has run and passed.

gavanderhoorn:
Thanks @cclauss :+1:

PR #469, opened 2019-11-25
Bdenoun/fix gazebo warning

bdenoun:
Gazebo 7 and 9 complains about a missing tag in the transmission file. Just added to avoid having warning messages in the logs


gavanderhoorn:
Thanks for contributing to these packages!

Issue #440, opened 2019-07-02
Suggested new repository structure


gavanderhoorn:
During WRID19 we (RIC-EU) have discussed the future of the packages in this repository and recongised the following challenges/constraints/requirements:

 1. the description, moveit and kinematics package(s) should probably stay together: changes to one often automatically need changes in the other(s), so having their release cycle synchronised makes sense
 1. `ur_driver` is deprecated and should not be used any longer
 1. `ur_bringup` ties everything to `ur_driver`, confuses users of other drivers and generally makes it difficult to use other drivers
 1. `ur_msgs` is currently used by `ur_modern_driver` *and* `ur_driver`, but is not necessarily tied to the state of any of the other packages. `ur_driver` is going to be retired, which leaves just `ur_modern_driver`. However, the new UR driver (FZI & UR) will also most likely also use `ur_msgs`
 1. `ur_gazebo` currently is almost empty (some `ros_control` config and `.launch` files), with most of the Gazebo-specific content actually hosted in `ur_description`

With this in mind, we propose the following new structure / division of packages:

 1. `ros-industrial/universal_robot`: hosts description, moveit, IK and gazebo packages
 1. `ros-industrial/ur_msgs`: hosts just the msgs package
 1. `ros-industrial/ur_modern_driver`: the modern driver itself, plus a "tools" package (ros-industrial/ur_modern_driver#282)
 1. the new UR & FZI driver (hosted at `http://github.com/UniversalRobots`)

Note that `ur_driver` and `ur_bringup` will no longer be retained in this proposed layout. The former is deprecated and the latter is tightly coupled with the former.

This setup allows us to keep packages that need to be in-sync, in-sync (descriptions, moveit and kinematics), separately version resources that don't need that (msgs), keep packages with high cohesion together (gazebo) while still separating them from the other pkgs and leave final bringup to the individual drivers.

gavanderhoorn:
@ipa-nhg: FYI.

@miguelprada also.

miguelprada:
I see one issue with the suggested layout.

We've been recently receiving a lot of issues by people getting confused about how to "connect" the MoveIt setup to the actual robot driver. With the proposed layout, this is very likely to continue being the case, since the correct MoveIt setup will depend on the topic _layout_ exposed by each specific driver.

[...]

PR #472, opened 2019-12-07
msgs: mark arch independent

gavanderhoorn:
As per subject.

As recommended by [REP-140](https://www.ros.org/reps/rep-0140.html#architecture-independent).


gavanderhoorn:
I'll re-open this after `ur_msgs` is migrated to its own repository (as discussed in #440, #448).

gavanderhoorn:
Replacement: https://github.com/ros-industrial/ur_msgs/pull/1.

PR #480, opened 2020-02-11
F#src 4095 fixing deprecation msgs

mikramarc:
Removing deprecated parameter

https://github.com/ros-planning/moveit/blob/melodic-devel/MIGRATION.md


gavanderhoorn:
Did you intend to submit this to the ShadowRobot fork instead?

mikramarc:
Hi @gavanderhoorn No. This parameter is deprecated in new moveit melodic version and is going to cause deprecation warnings, I assume you would like this change in your repository as well.

gavanderhoorn:
Thanks for submitting the PR in that case :100: 

However, I believe this is a duplicateof #474 in this case.

mikramarc:
True, I missed that, sorry about that :) Closing it.

gavanderhoorn:
No problem.

Thanks for submitting yours :)

PR #467, opened 2019-11-22
fix #466: fix loading e-series robot_description param

gonzalocasas:
Fixed the usages of `ur_description` (and matching urdf names) for all the e-series moveit_config packages to point to the correct `ur_e_description` package.

PR #481, opened 2020-02-17
Fix: Corrected references to ur_e in the three ur_e robots

### LINKED ISSUE #466: e-Series support: planning_context.launch loading the wrong robot
It looks like all 3 `planning_context.launch` files of the e-series moveit_config packages are pointing to the wrong robot. They all use `ur_description` instead of `ur_e_description` and point to the wrong file as well.

(It even looks like MoveIt is complaining about this during startup, with a warning about a mismatch between srdf and urdf)

tejaswid:
Similar to the commit fix #466 and commit 64c177f

### DIFF
--- a/ur10_e_moveit_config/launch/planning_context.launch
+++ b/ur10_e_moveit_config/launch/planning_context.launch
@@ -8,8 +8,8 @@
 
   <!-- Load universal robot description format (URDF) -->
   <group if="$(arg load_robot_description)">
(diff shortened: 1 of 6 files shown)


gavanderhoorn:
`melodic-devel` is currently catching up to `kinetic-devel`.

Please use `kinetic-devel` on Melodic instead.

gavanderhoorn:
As this was already fixed in #467, I'm going to close this.

Thanks for taking the time to create and submit the PR though :+1: much appreciated.

tejaswid:
Cool. No worries.

PR #500, opened 2020-06-09
Update Travis CI config

gavanderhoorn:
Changes:

 - don't allow Melodic jobs to fail
 - switch to `master` branch

Enabling `ccache` won't really help that much, as only `ur_kinematics` now still contains C++, but it won't hurt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
     - ROS_DISTRO="melodic" PRERELEASE=true
 
 install:
-  - git clone --quiet --depth=1 -b master https://github.com/ros-industrial/industrial_ci.git .ci_config
+  - git clone --quiet --depth=1 -b master https://github.com/ros-industrial/industrial_ci.git .industrial_ci
 
 script:
-  - .ci_config/travis.sh
+  - .industrial_ci/travis.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ git:
 env:
   global:
     - CCACHE_DIR=$HOME/.ccache
-    - ROS_REPO=ros
+    - ROS_REPO=main
   matrix:
     - ROS_DISTRO="kinetic"
     - ROS_DISTRO="kinetic" ROS_REPO=ros-testing
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,10 @@ env:
     - ROS_REPO=ros
   matrix:
     - ROS_DISTRO="kinetic"
-    - ROS_DISTRO="kinetic" ROS_REPO=ros-shadow-fixed
+    - ROS_DISTRO="kinetic" ROS_REPO=ros-testing
     - ROS_DISTRO="kinetic" PRERELEASE=true
     - ROS_DISTRO="melodic"
-    - ROS_DISTRO="melodic" ROS_REPO=ros-shadow-fixed
+    - ROS_DISTRO="melodic" ROS_REPO=ros-testing
     - ROS_DISTRO="melodic" PRERELEASE=true
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ env:
     - CCACHE_DIR=$HOME/.ccache
   matrix:
     - ROS_DISTRO="kinetic" ROS_REPO=main
-    - ROS_DISTRO="kinetic" ROS_REPO=ros-testing
+    - ROS_DISTRO="kinetic" ROS_REPO=testing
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
@ipa-mdl: if you could take a look and confirm I'm not doing anything stupid here, that would be appreciated.

mathias-luedtke:
> switch to master branch

Not necessary anymore :sunglasses:

gavanderhoorn:
> > switch to master branch
> 
> Not necessary anymore sunglasses

Still like to be explicit about this. It's very clear that this config is using `master`.

And we don't run the risk of you changing default branches on us ;)

.travis.yml
```diff
@@ -2,24 +2,27 @@ language: generic
...
-    - ROS_DISTRO="melodic"   ROS_REPO=ros
-    - ROS_DISTRO="melodic"   ROS_REPO=ros-shadow-fixed
-    - ROS_DISTRO="melodic"   PRERELEASE=true
-matrix:
-  allow_failures:
-    - env: ROS_DISTRO="melodic"   ROS_REPO=ros
-    - env: ROS_DISTRO="melodic"   ROS_REPO=ros-shadow-fixed
-    - env: ROS_DISTRO="melodic"   PRERELEASE=true
+    - ROS_DISTRO="kinetic"
+    - ROS_DISTRO="kinetic" ROS_REPO=ros-testing
```
mathias-luedtke: `ROS_REPO=[ros-]testing` is the default, you would run the same job twice.
To use the stable release, you have to set `ROS_REPO=ros` (even better: `ROS_REPO=main`) explicitly.
If you want to, you can mark the others with `ROS_REPO=testing` for readability.

.travis.yml
```diff
@@ -2,24 +2,27 @@ language: generic
...
+    - ROS_DISTRO="kinetic" PRERELEASE=true
+    - ROS_DISTRO="melodic"
+    - ROS_DISTRO="melodic" ROS_REPO=ros-testing
+    - ROS_DISTRO="melodic" PRERELEASE=true
+
 install:
-  - git clone --quiet --depth=1 https://github.com/ros-industrial/industrial_ci.git .ci_config
+  - git clone --quiet --depth=1 -b master https://github.com/ros-industrial/industrial_ci.git .ci_config
```
mathias-luedtke: Nitpick: I never understood why it was named `.ci_config`, I have updated all documentation to `.industrial_ci`.

.travis.yml
gavanderhoorn: Isn' that done by the `ROS_REPO=ros` in the `global` section?

mathias-luedtke: Ah yes, you are right. I did not see that line.
I would remove it as a global. You are adding both variants anyway.

gavanderhoorn:
Just noticed: this is the 500th ~~PR~~ PR or issue on this repository :clinking_glasses:

.travis.yml
```diff
@@ -2,24 +2,26 @@ language: generic
...
-    - ROS_DISTRO="melodic"   ROS_REPO=ros
-    - ROS_DISTRO="melodic"   ROS_REPO=ros-shadow-fixed
-    - ROS_DISTRO="melodic"   PRERELEASE=true
-matrix:
-  allow_failures:
-    - env: ROS_DISTRO="melodic"   ROS_REPO=ros
-    - env: ROS_DISTRO="melodic"   ROS_REPO=ros-shadow-fixed
-    - env: ROS_DISTRO="melodic"   PRERELEASE=true
+    - ROS_DISTRO="kinetic" ROS_REPO=main
+    - ROS_DISTRO="kinetic" ROS_REPO=ros-testing
```
mathias-luedtke: `ROS_REPO=testing` in the new naming convention..

`ros-testing` is the actually name of the ROS1 repository on the server..
For now it is treated like `testing`, but this might change in the future.

gavanderhoorn: I changed that based on a notice from `industrial_ci` in the first job that was run I believe.

Anyway, I'll change it.

mathias-luedtke: https://github.com/ros-industrial/industrial_ci/blame/master/industrial_ci/src/env.sh#L91 ?

mathias-luedtke (approved):

.travis.yml
gavanderhoorn: Probably, I cannot find it in the build log any more.

I must have misremembered.

gavanderhoorn:
Thanks for the review @ipa-mdl :+1:

PR #501, opened 2020-06-09
Cherry-pick missing commits from kinetic-devel into melodic-devel

gavanderhoorn:
As per subject.

This gets `melodic-devel-staging` to the same state (functionally) as current `kinetic-devel`.

`melodic-devel` was branched off of `kinetic-devel` at a point where releases `1.2.6` and `1.2.7` hadn't been made yet, so three commits were missing (see the [diff](https://github.com/ros-industrial/universal_robot/compare/melodic-devel...kinetic-devel)).

Rebasing `melodic-devel-staging` is not an option as it then cannot be merged with `melodic-devel`.

PRs 'merged' in this PR:

 - #368
 - #457 (partial, due to `ur_driver` having been removed entirely)
 - ~~#467 (the MoveIt configurations will be regenerated completely, but this completes their history)~~


gavanderhoorn:
Dropped the MoveIt e-Series commit as this will be fixed later, makes this PR fail and it makes little sense to fix that now just to pass CI.

PR #505, opened 2020-06-11
Remove incorrect base_link rotation.

### LINKED ISSUE #503: [ur_description] base meshes have wrong orientation
given the discussion in https://forum.universal-robots.com/t/dynamically-changing-gravity-vector/7633/6, 

```
[quote="felixmessmer, post:5, topic:7633, full:true"]
@Ebbe
so the wire is aligned with the positve y-axis of the gravity reference coordinate system?
[/quote]
[quote="Ebbe, post:6, topic:7633, full:true"]
Yes, the wire is aligned with y-axis of the robot base coordinate system.
[/quote]
```

it seems the meshes for the base (i.e. `base.dae` and `base.stl`) have wrong orientation...the cables are aligned with the negative y-axis of `base` link - but should be aligned with positive y-axis of `base` link

this affects only the non-e-series robots....e-series robots do not have a cable modeled...

the coordinate systems - mainly `base_link` and `base` - are correct, but I think the meshes should be rotated by 180°

---
can someone help me to verify this?
it's a bit of a brain twister with the various coordsystems in ROS and only semi-optimal display of axis on the TeachPanel

gavanderhoorn:
This fixes the long-standing problem with `base_link` being incorrectly rotated 180 degree over Z wrt the real robot (ie: "the connector was sticking out the wrong side").

This commit removes that rotation, while also making sure `base` keeps its correct location and orientation.

The `base->tool0` transform should be unaffected, as that had a 180 degree counter-rotation.

### DIFF
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -303,7 +303,12 @@
     <!-- ROS-Industrial 'base' frame: base_link to UR 'Base' Coordinates transform -->
     <link name="${prefix}base"/>
     <joint name="${prefix}base_link-base_fixed_joint" type="fixed">
-      <origin xyz="0 0 0" rpy="0 0 0"/>
+      <!-- Note the rotation over Z of pi radians: as base_link is REP-103
+           aligned (ie: has X+ forward, Y+ left and Z+ up), this is needed
+           to correctly align 'base' with the 'Base' coordinate system of
+           the UR controller.
+      -->
+      <origin xyz="0 0 0" rpy="0 0 ${pi}"/>
       <parent link="${prefix}base_link"/>
       <child link="${prefix}base"/>
     </joint>
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -76,7 +76,8 @@
     <xacro:ur_arm_transmission prefix="${prefix}" hw_interface="${transmission_hw_interface}" />
 
     <!-- links: main serial chain -->
-    <link name="${prefix}base_link" >
+    <link name="${prefix}base_link"/>
+    <link name="${prefix}base_link_inertia">
       <visual>
         <origin xyz="0 0 0" rpy="0 0 0"/>
         <geometry>
@@ -219,8 +220,13 @@
     </link>
 
     <!-- joints: main serial chain -->
-    <joint name="${prefix}shoulder_pan_joint" type="revolute">
+    <joint name="${prefix}base_link-base_link_inertia" type="fixed">
       <parent link="${prefix}base_link" />
+      <child link="${prefix}base_link_inertia" />
+      <origin xyz="0 0 0" rpy="0 0 0" />
+    </joint>
+    <joint name="${prefix}shoulder_pan_joint" type="revolute">
+      <parent link="${prefix}base_link_inertia" />
       <child link="${prefix}shoulder_link" />
       <origin xyz="${shoulder_x} ${shoulder_y} ${shoulder_z}" rpy="${shoulder_roll} ${shoulder_pitch} ${shoulder_yaw}" />
       <axis xyz="0 0 1" />
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -79,7 +79,7 @@
     <link name="${prefix}base_link"/>
     <link name="${prefix}base_link_inertia">
       <visual>
-        <origin xyz="0 0 0" rpy="0 0 0"/>
+        <origin xyz="0 0 0" rpy="0 0 ${pi}"/>
         <geometry>
           <mesh filename="${base_visual_mesh}"/>
         </geometry>
@@ -88,7 +88,7 @@
         </material>
       </visual>
       <collision>
-        <origin xyz="0 0 0" rpy="0 0 0"/>
+        <origin xyz="0 0 0" rpy="0 0 ${pi}"/>
         <geometry>
           <mesh filename="${base_collision_mesh}"/>
         </geometry>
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
@fmauch: could I ask you to verify this against an e-Series robot?

@fmessmer: this should resolve your #503.

gavanderhoorn:
A good test would be to run `rosrun tf2_tools echo.py base tool0` and comparing that with what the TP shows (for an all-zeros toolframe of course).

fmauch:
> @fmauch: could I ask you to verify this against an e-Series robot?

sure thing, I'll try to squeeze it in tomorrow.

fmauch:
I can confirm, that on our UR16e the TP and tf_echo `base->tool0` and `base->tool0_controller` all report the same values. :-)

fmauch (approved):

gavanderhoorn:
@fmauch wrote:
> I can confirm, that on our UR16e the TP and tf_echo `base->tool0` and `base->tool0_controller` all report the same values. :-)

Nice, thanks.

Seeing as the xacro macro is identical across all supported robot models, that should fix it for non e-Series as well.

gavanderhoorn:
While it's nice to have removed the rotation, the issue seems to be a bit deeper than I initially thought.

The current xacro macro has the robot stretched out over X- when in its zero-pose. While by convention (and REP-103), it should be pointing in the X+ direction.

In essence, the robot is now modelled *backwards*.

I'll see how I can best rectify this, but in any case it won't affect the `base->tool0` transform, as that is already correct and a few extra rotations of `base_link` will not change that (provided we counter-rotate `base`).

fmessmer:
well, while this PR aligns the cable bit of the `base` mesh with the y-axis of the `base` link - as it should be, it also changes the orientation of `base_link` which might cause trouble for people using the UR as part of a more complex scene or robot

yes, it is stated in the `macro` with a comment
and yes, there are breaking changes announced for melodic

but still, what I tried to bring up in #503 was way less intrusive, by simply adjusting the `origin` tag of `${prefix}base_link` (or rotate the meshes) - this would not have changed the tf tree

gavanderhoorn:
> well, while this PR aligns the cable bit of the `base` mesh with the y-axis of the `base` link - as it should be, it also changes the orientation of `base_link` which might cause trouble for people using the UR as part of a more complex scene or robot
> 
> yes, it is stated in the `macro` with a comment
> and yes, there are breaking changes announced for melodic

Exactly, so this is the point in time to do this.

I'm sorry for users who will have to update their models, but this is a `-devel` branch and nothing has ever been released for melodic.

> but still, what I tried to bring up in #503 was way less intrusive, by simply adjusting the `origin` tag of `${prefix}base_link` (or rotate the meshes) - this would not have changed the tf tree

The current model is wrong. It needs to be fixed.

gavanderhoorn:
Ok @fmessmer: I've used a different approach.

`base_link` remains as it is now (it was already REP-103 aligned), and I've introduced a new frame (`base_link_inertia`) which both serves to rotate things around to have X+ pointing "backward" (for the rest of the chain) as well as to deal with the `root-link-with-inertia` limitation in KDL.

Orientation of the `base` mesh is fixed with this, and the model is now a bit more coherent (although there are still rotations around Z of `pi` radians in several places).

Together with #506 I believe this now gives us a proper model, with the expected frames.

As always: `base->tool0` should return whatever the TP shows for that transform.

Let me know whether this works for you.

@fmauch: if you could check this *one last time* ?

fmauch:
And again, it reports the same values :-)

gavanderhoorn:
> And again, it reports the same values :-)

Good to hear :)

gavanderhoorn:
Travis is doing weird things again. Two PR builds apparently for one PR from one branch. And both have already completed, but the second one has not reported to GH it has completed ..

gavanderhoorn:
@fmessmer: I'm going to merge this by eod.

If you have any further comments not addressed then, we can fix those in subsequent PRs.

gavanderhoorn:
Just going to wait for Travis to be happy and then I'll merge this.

gavanderhoorn:
Thanks @fmauch for the early review and thanks @fmessmer for the additional input.

fmessmer:
sorry, for not getting back earlier....but the merged version works for me (only tested briefly)

zeal-up:
@gavanderhoorn 
Sorry, I still a little confusing about your patch.
In my understanding, the problems we should solves are: 1. make the cable point along the positive y axis. 2. the transfrom from tool0->base is the true tcp pose of the real robot(the pose read frm TP).

The introduction of the base_link_inertia frame is necessary, but even with your patch, the robot in rviz still has a rotation issue, it has a 180 degree rotation compare with the real robot.

So, why don't you align base_link and base frame? a.k,a, the joint between base and base_link is a fix joint with no rotation.
The fix procedures is as below(after your PR):
1. align base and base_link frame
```
<joint name="${prefix}base_link-base_fixed_joint" type="fixed">
      <!-- Note the rotation over Z of pi radians: as base_link is REP-103
           aligned (ie: has X+ forward, Y+ left and Z+ up), this is needed
           to correctly align 'base' with the 'Base' coordinate system of
           the UR controller.
      -->
      <origin xyz="0 0 0" rpy="0 0 0"/>
      <parent link="${prefix}base_link"/>
      <child link="${prefix}base"/>
    </joint>
```

2. set the rotation between base_link and base_link_inertia to be zero:
```
<joint name="${prefix}base_link-base_link_inertia" type="fixed">
      <parent link="${prefix}base_link" />
      <child link="${prefix}base_link_inertia" />
      <!-- 'base_link' is REP-103 aligned (so X+ forward), while the internal
           frames of the robot/controller have X+ pointing backwards.
           Use the joint between 'base_link' and 'base_link_inertia' (a dummy
           link/frame) to introduce the necessary rotation over Z (of pi rad).
      -->
      <origin xyz="0 0 0" rpy="0 0 0" />
    </joint>
```

All have be down.

Maybe I miss something, but it's pretty good in all my test:
1. the cable point along the positive y-axis
2. the tool0->base transform match the TP in the robot
3. the visualization is exactly match the real robot

PR #506, opened 2020-06-11
Add flange frame

gavanderhoorn:
As per subject.

This makes `ur_description` adhere to the (WIP) REP [Coordinate Frames for Serial Industrial Manipulators](http://gavanderhoorn.github.io/rep/rep-0199.html) and aligns it with other ROS-Industrial robot support packages (such as those for ABB and Fanuc).

Users will have to use `flange` as the EEF attachment point, instead of `ee_link`. `tool0` will always be coincident with an unconfigured (or "all zeros") toolframe, and must not be changed.

This is for #448.

### DIFF
--- a/ur_description/urdf/ur_macro.xacro
+++ b/ur_description/urdf/ur_macro.xacro
@@ -320,7 +320,8 @@
     <!-- Frame coincident with all-zeros TCP on UR controller -->
     <link name="${prefix}tool0"/>
     <joint name="${prefix}flange-tool0" type="fixed">
-      <origin xyz="0 0 0" rpy="0 0 0"/>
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
@fmauch: could I ask you to check the `base->tool0` transform corresponds to what the TP shows?

I've had to redo/include some more rotations and want to make sure it's all correct.

fmauch:
> @fmauch: could I ask you to check the `base->tool0` transform corresponds to what the TP shows?
> 
> I've had to redo/include some more rotations and want to make sure it's all correct.

Yep, seems to work properly.

gavanderhoorn:
Thanks @fmauch for the review :+1:

PR #511, opened 2020-06-16
Remove negative zeros

gavanderhoorn:
I'm assuming this is just a YAML serialisation artefact and those negative zeros don't actually need to be there @fmauch?


fmauch:
Yes, absolutely

fmauch (approved):
As assumed, they are simply from the yaml serialization.

gavanderhoorn:
Thanks @fmauch.

PR #513, opened 2020-06-16
Fix joint limits for all supported variants

gavanderhoorn:
I'm somewhat confused as to where the previous values for the limits came from, as there seemed to be quite some deviations from the specs (except for the UR16e, but that was only added very recently).

First commit fixes the limits (position, velocity and effort) based on the specs in the linked documents (user manuals and the support site article).

Second commit changes the `_macro.xacro` to accept joint limit data in degrees and degrees/sec (instead of radians). This makes it much easier to compare the limits in the files with the ones on the datasheets and the manuals.

Table used for joint torque limits (from [here](https://www.universal-robots.com/articles/ur-articles/max-joint-torques), but with torques instead of motor sizes):

```
Joint     UR3/UR3e     UR5/UR5e      UR10/UR10e    UR16e
Base      56 Nm (Sz2)  150 Nm (Sz3)  330 Nm (Sz4)  330 Nm (Sz4)
Shoulder  56 Nm (Sz2)  150 Nm (Sz3)  330 Nm (Sz4)  330 Nm (Sz4)
Elbow     28 Nm (Sz1)  150 Nm (Sz3)  150 Nm (Sz3)  150 Nm (Sz3)
Wrist 1   12 Nm (Sz0)   28 Nm (Sz1)   56 Nm (Sz2)   56 Nm (Sz2)
Wrist 2   12 Nm (Sz0)   28 Nm (Sz1)   56 Nm (Sz2)   56 Nm (Sz2)
Wrist 3   12 Nm (Sz0)   28 Nm (Sz1)   56 Nm (Sz2)   56 Nm (Sz2)
```

### DIFF
--- a/ur_description/config/ur10/joints_limits_parameters.yaml
+++ b/ur_description/config/ur10/joints_limits_parameters.yaml
@@ -1,14 +1,22 @@
 # Joints limits
+#
+# Sources:
+#
+#  - UR10 User Manual, Universal Robots, UR10/CB3, Version 3.13
+#    https://s3-eu-west-1.amazonaws.com/ur-support-site/69442/99203_UR10_User_Manual_en_Global.pdf
+#  - Support > Articles > UR articles > Max. joint torques
+#    https://www.universal-robots.com/articles/ur-articles/max-joint-torques
+#    retrieved: 2020-06-16, last modified: 2020-06-09
 limits:
   shoulder_pan:
     lower: -6.28318530718
     upper: 6.28318530718
-    velocity: 2.16
+    velocity: 2.0943951
     effort: 330.0
   shoulder_lift:
     lower: -6.28318530718
     upper: 6.28318530718
-    velocity: 2.16
+    velocity: 2.0943951
     effort: 330.0
   elbow_joint:
     # we artificially limit this joint to half it's actual joint limit to avoid
@@ -23,20 +31,20 @@ limits:
     # more information.
     lower: -3.14159265359
     upper: 3.14159265359
-    velocity: 3.15
+    velocity: 3.14159265359
     effort: 150.0
   wrist_1:
     lower: -6.28318530718
     upper: 6.28318530718
-    velocity: 3.2
-    effort: 54.0
+    velocity: 3.14159265359
+    effort: 56.0
   wrist_2:
     lower: -6.28318530718
     upper: 6.28318530718
-    velocity: 3.2
-    effort: 54.0
+    velocity: 3.14159265359
+    effort: 56.0
   wrist_3:
     lower: -6.28318530718
     upper: 6.28318530718
-    velocity: 3.2
-    effort: 54.0
+    velocity: 3.14159265359
+    effort: 56.0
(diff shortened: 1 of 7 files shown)


gavanderhoorn:
> I'm somewhat confused as to where the previous values for the limits came from, as there seemed to be quite some deviations from the specs

Because of this I'll look at the history of the various `.xacro`s and see whether there were any specific reasons for why the limits were so different from the (edit) *current* specs.

gavanderhoorn:
For the UR10 specifically, the incorrect joint limits (mostly velocity) were introduced in 78b27026.

Then changed again (to correct ones I believe) in 953d0465.

And then changed a few times more. At least for the UR10, it's a bit unclear where `2.16` (rad/s) comes from in 78b27026. 120 deg is 2.09 rad. A lower value than this could've made sense, but a higher value I don't understand.

---

Edit: ah. `urcontrol.conf` from a `v1.5.x` version of ursim specs the first joint of a UR10 as a `joint_size4_rev1.conf`. That file states:

```ini
[Joint]
min_limit = -6.2831853071795862
max_limit = 6.2831853071795862
v_max = 2.16 # 2.16 = 2/3 * 3.2
a_max = 60.0 # Guess
torque_max = 330
```
So `v_max` is `2.16` here. That's probably where the `2.16` in 78b27026 comes from.

Note also the `a_max` there :) Something which isn't documented any more (and probably incorrect for newer systems as well).

`v1.8.x` of ursim has the same data:

```ini
[Joint]
min_limit = -6.2831853071795862
max_limit = 6.2831853071795862
v_max = 2.16 # 2.16 = 2/3 * 3.2
a_max = 60.0 
torque_max = 330
```

Current manuals and datasheets spec this limit as 120 degree/sec though, which is `2.09` rad/s, lower than this `2.16`.

In the worst case, the maximum joint velocity planned with for a UR10 would be approx 3% slower than what the robot is capable of (although the `3.2` there is probably the actual maximum joint velocity).

gavanderhoorn:
I'm in doubt whether or not to keep 56fab86: while it's convenient to be able to directly compare the limits, it's also using non-standard units which goes against REP-103.

I'd like some additional input from others: @miguelprada @ipa-nhg @fmauch: thoughts?

fmauch:
> I'm in doubt whether or not to keep [56fab86](https://github.com/ros-industrial/universal_robot/commit/56fab86d26223ad5523ed6451f6985d19cca5a12): while it's convenient to be able to directly compare the limits, it's also using non-standard units which goes against REP-103.
> 
> I'd like some additional input from others: @miguelprada @ipa-nhg @fmauch: thoughts?

I think, going with radians (and maybe as a comment the according deg values) might be a good choice. As most of them are `PI` or `2PI`, it is still quite easy to compare to the specs.

gavanderhoorn:
True, but limiting them (artificially) is easier when it's in degrees and degrees/sec.

Using degrees will make re-using the limits files for MoveIt and `ros_control` essentially impossible though.

gavanderhoorn:
I'm going to remove 56fab86.

Radians is what it should be, and it allows reuse of the joint limits file for other packages/nodes.

If only `xacro` supported something like `subst_value` as `roslaunch` does.

gavanderhoorn:
I've checked and double-checked the limits for all variants and joints. I believe they are now correct.

If it turns out something is still not right, we'll fix that in a follow-up PR.

PR #516, opened 2020-06-17
Use 'ros_control style' joint limit files

gavanderhoorn:
As per subject.

This converts the joint limit files to 'ros_control style' and updates the xacro macro and `inc` files to be able to use those.

Converting these files to this format will allow us to reuse them in MoveIt configurations and for `ros_control` limits.

The current values are the ones updated in #513 (so the official ones), but users could of course copy one of these files, update them with preferred limits and then substitute the updated file by passing in the path to it.


gavanderhoorn:
Use with the `hardware_interface` in `ur_robot_driver` is currently blocked by UniversalRobots/Universal_Robots_ROS_Driver#198.

PR #518, opened 2020-06-24
Gazebo updates (extensive)

gavanderhoorn:
This is a draft PR (for now) to collect ongoing work towards updating the `ur_gazebo` package.

High-level goals:

 1. isolate Gazebo-specific infrastructure as much as possible so it does not 'leak' into any of the other packages
 1. treat Gazebo (and the simulated robot) as much as possible as just-another-robot, to allow (conceptual) reuse of workflows, `.launch` and configuration files from other packages

See the commit comments for rationales.

Mainly tested against Gazebo `9.x`.

This is -- and will be -- a rather substantial PR, but care has been taken to make sure commits make sense and reviews per-commit should be possible.

Usage is similar to `ur_robot_driver`: to start a UR10e with its kinematic calibration in Gazebo's empty world with a `JointStateController` and a `position_controllers/JointTrajectoryController` loaded and started:

```
roslaunch ur_gazebo ur10e_bringup.launch kinematics_config:=$(rospack find ur_calibration)/etc/ur10_example_calibration.yaml
```


gavanderhoorn:
CI is failing as `roslaunch` in Kinetic does not have ros/ros_comm#1520.

I'm going to ignore the Kinetic failures, as we don't target it with this refactor.

I'll probably update CI to allow Kinetic builds to fail (but still try to not break things on Kinetic for no reason).

gavanderhoorn:
@ipa-hsd: would you be willing to take a look at this?

Not just whether it works, but I'm also wondering whether the structure and distribution of files make sense to ppl.

And I'm curious how straightforward it would be to reuse the Gazebo description in other simulations. One thing that will be problematic is the 'hard-coded' `gazebo_ros_control` plugin. I'm thinking of making that configurable somehow, but I'm not sure I want to add yet *another* argument to the `_macro.xacro`.

gavanderhoorn:
Added a few commits which improve the "public interface" of `ur_gazebo`:

 - load the URDF and start the RSP as part of the bringup `.launch` file (instead of the `ros_control` part): this should make it easier for users to load custom kinematics configurations, as well as joint limits and altered visual appearances (as the arguments are available directly at the level of the bringup `.launch` file)
 - hide all non-entry-point `.launch` files by giving them a `.xml` suffix: this prevents them from showing up in `roslaunch` tab-completion

Also use the `dirname` substitution arg to avoid crawling the ROS pkg path for local include files.

---

Edit: `dirname` is not available on Kinetic, so this increases breakage with Kinetic.

gavanderhoorn:
@ipa-hsd: if you've already pulled, you may want to update, as these updates improve on some issues I encountered while trying to integrate the models provided by `ur_gazebo` with an existing simulation.

gavanderhoorn:
@fmauch: I'd appreciate it if you could take a look as well.

In essence I've tried to mimic the way `ur_robot_driver` structures its `.launch` files (see the OP for rationale).

Related to https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/97#discussion_r441717657: which parameters/arguments should we/do we want to expose?

gavanderhoorn:
I'm re-arranging some files again, so a proper review would not make sense yet @ipa-hsd and @fmauch (but then this is also still just a draft).

gavanderhoorn:
@ipa-hsd @fmauch: I'm pretty happy with the current state of this, so if you could take a look I'd appreciate it.

I will be adding one or two commits, but those will mostly *add* things which weren't there before, and will be largely orthogonal to the commits already here.

hsd-dev:
@gavanderhoorn sorry about the delay. I went through the PR and the commits this morning. And it works as expected. For me, it is a go :+1:

gavanderhoorn:
In order for all of this to really make sense, we should really include a set of convenience `_macro.xacro`s which provide default paths to the various per-variant `.yaml` files in `ur_description/config`. Right now even just including the default `ur_robot_gazebo` will require users to provide values for all 4 `*_parameters_files` parameters, which is rather annoying.

I won't do that in this PR though, as it's big enough, and requires changes in `ur_description` as well.

gavanderhoorn:
@ipa-hsd wrote:
> @gavanderhoorn sorry about the delay. I went through the PR and the commits this morning. And it works as expected. For me, it is a go +1

thanks @ipa-hsd.

I'll wait a bit for @fmauch's review and then I'll merge.

gavanderhoorn:
And something to consider for a later PR: should `ur_gazebo` be switched to `hardware_interface/EffortJointInterface`?

Right now a `hardware_interface/PositionJointInterface` is used, but this has several disadvantages (interaction with objects in the rest of the simulation is really bad fi).

@ipa-hsd @ipa-nhg @miguelprada: thoughts?

gavanderhoorn:
@fmauch: I've just force-pushed the latest version (rebased on-top of #520).

Would you perhaps have time to take a look today or tomorrow?

I'm mostly interested to know whether someone with a fresh eye can make sense of the structure of `ur_gazebo`, or whether it's completely illogical.

Note: I will add a more extensive readme to the package to explain which files should be used for which purpose and in which case (ie: top-level xacro should not be used in case a UR is to be integrated into a larger scene, etc).

fmauch:
> Would you perhaps have time to take a look today or tomorrow?

I'll try to fit it in this afternoon / evening.

ipa-nhg (approved):
Tested on:
- Ubuntu 18.04
- Rosdistro: Melodic
- Gazebo version: 9.0.0

I started the launch for the simulation and tested the controllers using the rqt_joint_trajectory_controller plugin, it works as expected.

General review: I checked the filesystem structure and it makes completely sense to me.

@gavanderhoorn Thanks for the contribution

gavanderhoorn:
Thanks for the review @ipa-nhg 👍 

Would you have any thoughts on this:

> And something to consider for a later PR: should `ur_gazebo` be switched to `hardware_interface/EffortJointInterface`?
> 
> Right now a `hardware_interface/PositionJointInterface` is used, but this has several disadvantages (interaction with objects in the rest of the simulation is really bad fi).

Perhaps I should post a new issue about it, so as to not discuss this in the comments under a PR review.

---

Edit: I've posted #521 for the discussion about the default `hardware_interface`.

ipa-nhg:
@gavanderhoorn As you suggest, we should open a new issue for the controllers type that is a perfect candidate to be solved during the WRID20. In principle, I am agree with the change but I want to evaluate and test it separately.

ur_gazebo/launch/inc/load_ur.launch.xml
```diff
@@ -0,0 +1,43 @@
...
+  -->
+
+  <!--Parameter files -->
+  <arg name="joint_limit_params" doc="YAML file containing the joint limit values"/>
+  <arg name="kinematics_params" doc="YAML file containing the robot's kinematic parameters. These will be different for each robot as they contain the robot's calibration."/>
+  <arg name="physical_params" doc="YAML file containing the phycical parameters of the robots"/>
+  <arg name="visual_params" doc="YAML file containing the visual model of the robots"/>
+
+  <!--Common parameters  -->
+  <arg name="transmission_hw_interface" default="hardware_interface/PositionJointInterface" />
```
fmauch: Is there no doc field on this on purpose, as this is rather common?

ur_gazebo/launch/inc/load_ur.launch.xml
```diff
@@ -0,0 +1,43 @@
...
+  <arg name="safety_limits" default="false" doc="If True, enable the safety limits controller"/>
+  <arg name="safety_pos_margin" default="0.15" doc="The lower/upper limits in the safety controller" />
+  <arg name="safety_k_position" default="20" doc="Used to set k position in the safety controller" />
+
+  <param name="robot_description" command="$(find xacro)/xacro '$(find ur_gazebo)/urdf/ur.xacro' --inorder
```
fmauch: as we are targeting melodic, `--inorder` could be omitted, right?

ur_gazebo/launch/inc/ur_control.launch.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0"?>
+<launch>
+  <!--
+    This file 'pretends' to load a driver for a UR robot, by accepting similar
```
fmauch, line 4: :+1: for the nice description!

ur_gazebo/launch/inc/ur_control.launch.xml
```diff
@@ -0,0 +1,59 @@
...
+  <arg name="controllers" default="joint_state_controller pos_joint_traj_controller"/>
+  <arg name="stopped_controllers" default="joint_group_pos_controller"/>
+
+  <!-- Gazebo parameters -->
+  <arg name="gazebo_model_name" default="robot" doc="The name to give to the model in Gazebo (after spawning it)." />
+  <arg name="gazebo_world" default="worlds/empty.world" doc="The '.world' file to load in Gazebo." />
+  <arg name="gui" default="true" doc="If true, Gazebo UI is started. If false, only start Gazebo server." />
+  <arg name="paused" default="false" doc="If true, start Gazebo in paused mode. If false, start simulation as soon as Gazebo has loaded." />
+  <arg name="robot_description_param_name" default="robot_description" doc="Name of the parameter which contains the robot description (ie: URDF) which should be spawned into Gazebo." />
+  <arg name="spawn_z" default="0.1" doc="At which height the model should be spawned." />
```
fmauch: Maybe a note that setting this to 0 could have negative consequences could reduce support questions?

fmauch (commented):
Also looks really clean to me. I only had a couple of small suggestions, but nothing critical.

Tested against gazebo 11 using ros-melodic and it seems to work finr.

ur_gazebo/launch/inc/load_ur.launch.xml
gavanderhoorn: Simply an oversight.

ur_gazebo/launch/inc/load_ur.launch.xml
gavanderhoorn: Yes.

I had hoped to maintain Kinetic compatibility, but some of Melodic's functionality is too convenient.

ur_gazebo/launch/inc/ur_control.launch.xml
gavanderhoorn: Done.

Although I don't expect ppl to actually *read* these :)

gavanderhoorn:
Ok.

Given the nr of reviews, the green CI and the state of this I'm going to merge the PR.

It's not perfect yet, but I believe it's a good set of changes which gets us where we want to go with `ur_gazebo`.

Further improvements and fixes can be done in follow-up PRs.

gavanderhoorn:
Thanks for the reviews @ipa-hsd @ipa-nhg and @fmauch :+1:

PR #520, opened 2020-06-30
Load yaml files in read_model_data(..), nowhere else

gavanderhoorn:
As per subject.

The current implementation accepts parameters with names such as `joint_limits_parameters_file`, but those parameters are (implicitly) assumed to contain the *contents* of the *joint_limits_parameters* file (ie: the YAML), instead of being actual (absolute) paths to that file.

This was confusing (at least to me).

To avoid having to `yaml_load(..)` in multiple places (basically every user of `read_model_data(..)` had to `yaml_load(..)` those files), this PR changes `read_model_data(..)` to invoke `yaml_load(..)` instead.

This also means the names of those parameters now convey what they should contain (ie: paths, not file contents).


gavanderhoorn:
Note: files in `ur_gazebo` are also updated, which will conflict with #518.

#518 will be rebased after this one has been merged.

PR #524, opened 2020-07-07
Fix ur3 visual mesh

### LINKED ISSUE #425: Bad quality visual meshes for UR3 in ros Kinetic
image

Visual meshes are bad quality for UR3 in ros kinetic especially the wrist3.dae. 

The mesh is not even centered when moving it with the joint_state_puplisher.

image

Other meshes have strange shading and are not smooth.

image

Briancbn:
Fix #425 
For Wrist3 Node 1 (ring) I simply added a edge split modifier and shade smooth
For Wrist3 Node 0 (end-effector) I rebuild the mesh in blender (maybe because I am not good enough or the resolution of the STL is too low, I don't see the screw holes coming back easily with edge split). It was drawn on top of the original STL so there should be no dimensional problem here.
For Wrist2 and Wrist1 blue logo cap, I use the UR3e ones, remove the screws. The original mesh will takes a very long time to fix (maybe I am not that good). Since I cannot regenerate STL

You can cherry-pick this commit onto `kinetic-devel` branch

Before:
Before

After:
after

Final:
Final


gavanderhoorn:
The meshes look really nice @Briancbn.

Could you comment on the file size difference?

Good looking meshes are important, but if it increases the size of the files too much (Collada can get very verbose), we may need to tweak them a little.

Briancbn:
Before
716K	wrist1.dae
724K	wrist2.dae
44K	        wrist3.dae

After
1.3M	wrist1.dae
1.2M	wrist2.dae
388K	wrist3.dae

I can try to reduce wrist3 if you think the differences are too big.

ipa-nhg (approved):
@Briancbn thanks for this contribution! Great job, I really like the new visualization.

I am ok with the new sizes. I let @gavanderhoorn to give the last :+1: and merge this

gavanderhoorn:
While this certainly improves the smoothness of the mesh, it increases their size 2x (wrist 3 about 9 times).

Personally I feel that's a bit too much, but I'm just one voice here.

It seems @ipa-nhg is ok with it, so it's a tie.

Perhaps someone else can weigh in.

Briancbn:
I will try to reduce the wrist 3 mesh

gavanderhoorn:
The other two may be more efficient to reduce. They are the main cause of the size increase I believe.

Briancbn:
> For Wrist2 and Wrist1 blue logo cap, I use the UR3e ones, remove the screws. The original mesh will takes a very long time to fix (maybe I am not that good). Since I cannot regenerate STL

@gavanderhoorn For the other 2, it is pretty much the best I can do here based on my ability. As the automatic generated meshes are not clean (nice quad for instance), it is pretty hard to directly use "decimate".

I assume as it was not an issue in ur3e, it could still be ok here.

Issue #521, opened 2020-07-01
Change ur_gazebo to EffortJointInterface [discussion]


gavanderhoorn:
As part of the `melodic-devel-staging` breaking changes (#448) we should consider whether to change the models in `ur_gazebo` to a `hardware_interface/EffortJointInterface` by default, instead of the `hardware_interface/PositionJointInterface` they have now.

Gazebo works better when everything in a simulation is (ultimately) effort controlled, especially when models in the simulation need to interact.

While the updated `ur_gazebo` (#518) allows for the `hardware_interface` to be configured using a `.launch` file, making use of any `hardware_interface` other than `PositionJointInterface` will also require updates to the controllers (which are currently configured to expect a `PositionJointInterface`).

From a UX perspective, only a whole-sale change from position to effort makes sense to me.

A disadvantage would be that the effort interface requires properly tuned PIDs to work correctly *and* have the `inertia` and `mass` parameters properly set.

And this means PIDs for the no-load case as well as the case where the robot is outfitted with a gripper and is lifting some object.

The latter could be left to the user/integrator, but the former would certainly have to be done by the maintainers here.

gavanderhoorn:
Related previous discussion: #87 and #161 (which actually removed the `EffortJointInterface`) and #227 (which makes it configurable again).

gavanderhoorn:
We could also configure a set of both position *and* effort controllers in the controller configuration `.yaml`s.

The position ones will be started by default, with the option for the user to switch to the effort ones.

It would be the user's responsibility to make sure she's also started the simulation with the effort `hardware_interface`.

Trying to load/start an effort-based controller when that interface is unavailable will simply return a nr of fatal errors.

matteolucchi:
First: thanks for the great work @gavanderhoorn, your contributions have been very helpful ! 

I am also struggling with some issues related to the use of the `PositionJointInterface`. I will try to tune the PID params, at least for the UR10 and if I have any success I'll report back here

gavanderhoorn:
@matteolucchi wrote:
> I am also struggling with some issues related to the use of the `PositionJointInterface`. I will try to tune the PID params, at least for the UR10 and if I have any success I'll report back here

Please make sure to check the `melodic-devel-staging` branch, as things may have changed wrt `kinetic-devel`.

And #518 changes quite a few things in `ur_gazebo`, so that would be something to take into account as well.

[...]

PR #504, opened 2020-06-11
Updated cog for ur16 how I understand the documentation

fmauch:
This PR spins out of #460, as it not only affects the UR16e.

Up until now, the robots' CoG calculation were centered inside the links they were referring to. This PR sets them to the place where they are supposed to be according to [UR's documentation](https://www.universal-robots.com/how-tos-and-faqs/faq/ur-faq/parameters-for-calculations-of-kinematics-and-dynamics-45257/)

The actual inertia matrix isn't changed by this PR, that might still have to be updated. UR provides 3x3 inertia matrices, but (at least in URSim) the ones from the ur3e and ur5e seem to be very sparse (the following files are from a URSim v5.8, so all robots are actually e-Series):
```
urcontrol.conf.UR3:inertia_matrix = [ [0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0, 0, 0, 0.000144] ]
urcontrol.conf.UR16:inertia_matrix = [ [0.03351, 0.00002, -0.00001, 0.00002, 0.03374, 0.00374, -0.00001, 0.00374, 0.02100], [0.02796, -0.00010, -0.00720, -0.00010, 0.47558, 0.00003, -0.00720, -0.00001, 0.47635], [0.01091, 0.00006, 0.01012, 0.00006, 0.12060, 0.00001, 0.01012, 0.00001, 0.11714], [0.00609, -0.00001, -0.00000, -0.00001, 0.00245, 0.00083, 0.00000, 0.00083, 0.00579], [0.00389, -0.00001, 0.00000, -0.00001, 0.00219, -0.00045, 0.00000, -0.00045, 0.00363], [0.00117, 0.00000,  0.00000, 0.00000, 0.00118, 0.00000, 0.00000, 0.00000, 0.00084] ]
[...]

Issue #386, opened 2018-10-11
UR10 Gazebo Launching Problem


mjpatell:
Hello all,
I am using UR10 arm. While lunching gazebo model of UR10 arm in the kinetic, it is falling down on the ground. I think but not sure this is the problem of the gravity. I tried to tune this parameter but not successed yet. If any one facing same problem and solve it than please let me know.

My setup:

- Ubuntu 16.04
- ROS Kinetic
- UR10 arm

shloakmehta:
gazebo 7 or 9?

mjpatell:
gazebo 7

shloakmehta:
So I get these errors too which - according to the last two warnings - gravity acts weird, and to solve it, we either have to  set gazebo_pid parameters, switch to the VelocityJointInterface or EffortJointInterface, or upgrade to Gazebo 9.

I may be wrong and am about to try myself right now with the UR10 but according to https://github.com/ros-simulation/gazebo_ros_pkgs/issues/612 it should be the solution. 

Let me know how you get on...
```
[ERROR] [1542388350.763274103, 0.342000000]: No pgain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/shoulder_pan_joint
[ERROR] [1542388350.765508662, 0.342000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/shoulder_lift_joint
[ERROR] [1542388350.767169872, 0.342000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/elbow_joint
[ERROR] [1542388350.768819330, 0.342000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/wrist_1_joint
[ERROR] [1542388350.770428920, 0.342000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/wrist_2_joint
[ERROR] [1542388350.772307131, 0.342000000]: No p gain specified for pid.  Namespace: /gazebo_ros_control/pid_gains/wrist_3_joint
[ INFO] [1542388350.785158214, 0.342000000]: Loaded gazebo_ros_control.
[ WARN] [1542388350.786564225, 0.343000000]: The default_robot_hw_sim plugin is using the Joint::SetPosition method without preserving the link velocity.
[ WARN] [1542388350.786622491, 0.343000000]: As a result, gravity will not be simulated correctly for your model.
[...]

PR #525, opened 2020-07-07
Switch UR10, UR10e and UR16e to effort controllers

ipa-kut:
This fix addresses https://github.com/ros-industrial/universal_robot/issues/521 which in turn addresses https://github.com/ros-industrial/universal_robot/issues/386, and was worked on during WRID2020

Based on a rough understanding of [this comment](https://github.com/ros-industrial/universal_robot/pull/504#issuecomment-652394378), only the three URs mentioned in the title have been updated.

PID values for the effort controllers were taken from [here](https://github.com/fmauch/universal_robot/commit/0d2ec11e2ac4a55317ffe427e16c1a949a1a4bee) 

After making these changes, the gazebo simulation of the robot was observed to start in a "weird" pose :
UR16e:   
initial_pose_16e

UR10 and UR10e:   
initial_pos_10e

However, the controllers initially seem to initially work well, and the robots could be jogged using the [RQT Joint trajectory controller plugin](http://wiki.ros.org/rqt_joint_trajectory_controller)

### DIFF
--- a/ur_gazebo/config/ur10_controllers.yaml
+++ b/ur_gazebo/config/ur10_controllers.yaml
@@ -2,8 +2,8 @@ joint_state_controller:
   type: joint_state_controller/JointStateController
   publish_rate: &loop_hz 125
 
-pos_joint_traj_controller:
-  type: position_controllers/JointTrajectoryController
+eff_joint_traj_controller:
+  type: effort_controllers/JointTrajectoryController
   joints: &robot_joints
     - shoulder_pan_joint
     - shoulder_lift_joint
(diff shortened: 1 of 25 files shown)


fmauch:
As this PR is branching of `melodic-devel-staging` none of the models use the updated cog from #504.

Eventually, this should be changed for all 7 robot variants, though.

ipa-kut:
I will change this for the remaining URs now.

> As this PR is branching of `melodic-devel-staging` none of the models use the updated cog from #504.

Would it be better to wait until #504 is merged, to rebase this and test again?

fmauch:
> I will change this for the remaining URs now.
> 
> > As this PR is branching of `melodic-devel-staging` none of the models use the updated cog from #504.
> 
> Would it be better to wait until #504 is merged, to rebase this and test again?

I think, this can be implemented independently from #504 but for fine tuning it might be interesting whether it makes a difference whether we use the cylinder approximation for the inertia or we use the matrices provided by the robot controller. So, testing and fine-tuning this might require a rebase on #504. However, as the results from those tests will also impact #504, we should not wait for that to be merged, but develop them in parallel. I think, this won't be too much of an issue as there are no conflicts between those two.

ipa-nhg (approved):
Tested with https://github.com/ros-industrial/universal_robot/pull/526 and it works.

Great contribution @ipa-kut, thanks! :+1:

RobertWilbrandt:
The changes from this should now be merged through #619. Thanks @ipa-kut for getting this started!

PR #528, opened 2020-07-14
Add transmission_hw_interface to URe xacro and expose everywhere

pcw109550:
Adds the capability to select the hardware interface used in the `<transmission>` elements in the URDF model. This allows creating custom simulation setups with different hardware interfaces, e.g. `hardware_interface/VelocityJointInterface`.

Same changes compared with #392, applying to all URe models in this case.


ipa-nhg (approved):
Thanks  a lot for this contribution, I think the changes make sense for this branch and for consistency with #392 .

(Comment: the current version of this repository (for melodic), which have common launch files for both UR series, supports already this feature: https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur_gazebo/launch/inc/load_ur.launch.xml#L28)

Issue #430, opened 2019-06-12
Update MoveIt configurations


gavanderhoorn:
Even though the provided MoveIt configurations are only there for convenience and quick testing of new robot setups (ie: right after unboxing) and are not intended for longer use (as any real application would need a customised MoveIt configuration), they should be up-to-date with the latest MoveIt changes.

This includes planner configurations, parameter and launch file names and structure of the package.

Rerunning the MSA and regenerating the packages would be a good first step, but care must be taken to not accidentaly override important configuration or remove settings that are needed for the correct functioning of MoveIt and UR robots.

gavanderhoorn:
Related: #431.

jacknlliu:
@gavanderhoorn just do it, please!

gavanderhoorn:
@jacknlliu: these issues were added in preparation for World ROS-I Day 2019.

Note the `support level: community` in the readme of this repository.

If you participate, perhaps you could even do it!

jeritgeorge:
I can work on this

gavanderhoorn:
@jeritgeorge: nice.

Please first investigate the current layout of the package. There are a few non-standard things that the UR moveit packages have.

It might be insightful to generate a completely new MoveIt configuration in a separate location and then comparing that to the current ones.

gavanderhoorn:
Did you make any progress here @jeritgeorge?

jeritgeorge:
Yes, I've reran the MSA for the ur3 config and went through and compared it to the current one. I think I have a decent understanding of the current setup and should at least have the ur3 config done either tonight or tomorrow. I think once one is done, the rest should be relatively simple.

gavanderhoorn:
I'm going to unassign this @jeritgeorge.

You're welcome to keep working on it of course.

fmauch:
I'l mark this closed, as #538 is merged.

Issue #431, opened 2019-06-12
Align MoveIt launch files with other ROS-Industrial packages


gavanderhoorn:
The MoveIt packages in this repository use a naming convention that is different from other ROS-Industrial repositories.

This reduces the reusability of experience with other packages and requires users to update launch files whenever using a different robot model.

It would be good to restructure the lauch files such that they conform to conventions used in other repositories (such as Fanuc, ABB and Motoman).

gavanderhoorn:
Related: #430.

fmauch:
This has been included into #538, so this should be good to close.

PR #538, opened 2020-09-24
Update MoveIt! support

### LINKED ISSUE #494: No planning library
Hi,
I have installed ros-melodic on a VM iwth Ubuntu 18.04 and built the universal_robot package from source as described in the repository https://github.com/ros-industrial/universal_robot.
I already tried launching the joint limited version of ur5 but I still get the RuntimeError: Unable to connect to move_group action server 'move_group' within allotted time (30s). And I assume this causes that there is no planning library loaded.

I am running the following commands, each in a different command window:

roslaunch ur_gazebo ur5.launch limited:=true

roslaunch ur5_moveit_config ur5_moveit_planning_execution.launch sim:=true limited:=true

roslaunch ur5_moveit_config moveit_rviz.launch config:=true

### LINKED ISSUE #573: this package is outdated with calibration_devel branch
Hi, as you all know, this package is outdated with 100+ commit with https://github.com/fmauch/universal_robot/tree/calibration_devel
This is confusing and breaks a lot of things. Why not just synchronize all commits from Felix?

fmauch:
MoveIt! support has not been migrated to the new description structure. So, this PR aims at implementing #430.

As raised in https://github.com/ros-industrial/universal_robot/issues/448#issuecomment-644570961 this is a first draft version find out what the desired state should be.

Basically, I recreated the package from scratch using the MSA and replaced the files in the existing package.

I removed all gazebo references and straightened a couple of minor issues.

Things I would like to specifically discuss about:
 - Our current xacro file generates a robot named "ur". Therefore, the moveit_config also contains files for `ur_<...>` instead of `ur16e_<...>` which I adapted manually in most cases. It might be a good idea to add the actual robot name as a parameter to [`ur.xacro`](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur_description/urdf/ur.xacro)
 - How should we handle the default expected controller? In the old configuration the controller name [was set to `""`](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur10_e_moveit_config/config/controllers.yaml#L2) effectively waiting for an action server coming up at `/follow_joint_trajectory`. This raised a lot of issues in the `ur_robot_driver` about people expecting this to work together out of the box. I left this at `manipulator` for now as a reminder that we should resolve this. In my local setup I set it to `scaled_pos_joint_traj_controller` in order to work with the `ur_robot_driver` out of the box.
 - The file `ur10_e_moveit_planning_execution.launch` was not created by the MSA. I don't know whether I did something wrong or whether this has changed nowadays.


Victorlouisdg:
Hi, I was wondering how you generate the MoveIt config files, do you have to manually walk through the MSA GUI for each UR robot? Is there a fixed procedure? Also, what is the .setup_assistant file used for?

fmauch:
> Hi, I was wondering how you generate the MoveIt config files, do you have to manually walk through the MSA GUI for each UR robot? Is there a fixed procedure? Also, what is the .setup_assistant file used for?

I currently created the files using the MSA GUI and after that did some manual modifications as mentioned above. From my point of view, all other moveit configurations could be generated by this one using search&replace mechanisms once we consented on a final version of this package. @gavanderhoorn  can probably say more about this.

fmauch:
@gavanderhoorn Any comments on this, so I can push this forward?

lianghongzhuo:
The urdf files for the robot are probably missing in the melodic-devel-staging branch. As a result, the moveit config you edited here cannot find a robot description file. Any plan to push this branch forward?

fmauch:
> The urdf files for the robot are probably missing in the melodic-devel-staging branch. As a result, the moveit config you edited here cannot find a robot description file. Any plan to push this branch forward?

If you are referring to the failing tests this is because of the tests not being updated. This is also related to the third of my questions above.

ipa-nhg:
@fmauch  Thanks a lot for this contribution, I made a quick review of the changes and it looks good to me! :+1: 

 A test on a the real hardware is pending!

Some answers/opinions to your questions:

> * Our current xacro file generates a robot named "ur". Therefore, the moveit_config also contains files for `ur_<...>` instead of `ur16e_<...>` which I adapted manually in most cases. It might be a good idea to add the actual robot name as a parameter to [`ur.xacro`](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur_description/urdf/ur.xacro)

I am agree, we should update the ur.xacro to get the robot "name" as parameter. But we need a consensus here, @gavanderhoorn what do you think?


> * How should we handle the default expected controller? In the old configuration the controller name [was set to `""`](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur10_e_moveit_config/config/controllers.yaml#L2) effectively waiting for an action server coming up at `/follow_joint_trajectory`. This raised a lot of issues in the `ur_robot_driver` about people expecting this to work together out of the box. I left this at `manipulator` for now as a reminder that we should resolve this. In my local setup I set it to `scaled_pos_joint_traj_controller` in order to work with the `ur_robot_driver` out of the box.

I think for the simulation the `follow_joint_trajectory` action is mapped to `/arm_controller/follow_joint_trajectory` . I am not sure if there is a clean way to set the namespace as argument of a top launch file, then should be ok if we keep the name empty. 

> * The file `ur10_e_moveit_planning_execution.launch` was not created by the MSA. I don't know whether I did something wrong or whether this has changed nowadays.

If you check the  [`ur10_e_moveit_planning_execution.launch`](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur10_e_moveit_config/launch/ur10_e_moveit_planning_execution.launch) it just include the `move_group.launch` and the file is only used by the test, then I vote for modifying the [test file](https://github.com/ros-industrial/universal_robot/blob/melodic-devel-staging/ur10_e_moveit_config/tests/moveit_planning_execution.xml) to include directly move_group. NOTE: For the simulation case the remap of the follow_joint_trajectory controller is required.

fmauch:
> I am agree, we should update the ur.xacro to get the robot "name" as parameter. But we need a consensus here, @gavanderhoorn what do you think?

As @gavanderhoorn mentioned in https://github.com/ros-industrial/universal_robot/issues/556#issuecomment-737158075 he already prepared a commit for this in https://github.com/gavanderhoorn/universal_robot/commit/4187b12e7318d626cc179d634094214c9a8368b9.

fmauch:
I've updated the PR building ontop of the changes from @gavanderhoorn mentioned above. It might be cleaner to merge the description changes in a separate PR first and then rebase this ontop of that. (**Edit**: Especially, since those changes break other packages like ur_gazebo)

If no objections rise, I can make this a full PR asap, so we can start the real review process (although other robot models will be pretty much copy&paste so reviewing one robot model might be simpler. What do you think @ipa-nhg?

ipa-nhg:
I am agree, we have to merge first the description changes.

You can go ahead with the rest of the robots. I made today a quick test on simulation and the planner worked as expected 👍🏽 

Test commands:

```
roslaunch ur_gazebo ur10_bringup.launch
roslaunch ur10_moveit_config move_group.launch
```

Then using the MoveIt! RVIZ plugin  I gave different goal positions, also adding a ground and some boxes as collisions, and the planner was able to find and execute trajectories without problems. 

I made a couple of minor changes to make this version working on simulation: https://github.com/fmauch/universal_robot/compare/moveit...ipa-nhg:moveit_sim_test?expand=1

fmauch:
Thanks for testing with gazebo. I did a test with URSim and the `ur_robot_driver` and it also worked as expected.

fmauch:
@ipa-nhg Concerning the changes you needed to make to `move_group.launch` it might be beneficial to indeed keep the `ur10_e_moveit_planning_execution.launch` file, so it can be used with both, a real robot and a gazebo robot.

fmauch:
I went ahead and finished the moveit configurations for all robots. The branch will be cleaned up once #562 is merged.

A couple of remarks to decisions I took on the way:
 - I set the default controller of `ur_robot_driver` into the default controller for the moveit configurations to enable users an easy start with the real robot.
 - I've renamed the e-series packages so that they match to the notation we use everywhere else (ur10e instead of ur10_e)

For the record: All packages are copied from the `ur10_moveit_config` package using the following commands inside the copied package:
```bash
export ROBOT=ur10e # just as an example
find . -type f -print0 | xargs -0 sed -i -e "s/ur10/${ROBOT}/g"
mv ./config/ur10.srdf ./config/${ROBOT}.srdf && \
  mv ./launch/ur10_moveit_controller_manager.launch.xml ./launch/${ROBOT}_moveit_controller_manager.launch.xml && \
  mv ./launch/ur10_robot_moveit_sensor_manager.launch.xml ./launch/${ROBOT}_robot_moveit_sensor_manager.launch.xml && \
  mv ./launch/ur10_robot_moveit_controller_manager.launch.xml ./launch/${ROBOT}_robot_moveit_controller_manager.launch.xml && \
  mv ./launch/ur10_moveit_planning_execution.launch ./launch/${ROBOT}_moveit_planning_execution.launch && \
  mv ./launch/ur10_moveit_sensor_manager.launch.xml ./launch/${ROBOT}_moveit_sensor_manager.launch.xml
```

As of writing this: The joint limits should probably be changed inside the individual packages, so that remains a ToDo.

**Edit:** Updated joint limits in a4237e8

gavanderhoorn:
I'd like to merge this @fmauch, but this has conflicts because #562 was just merged.

fmauch:
I'll update accordingly, thanks for letting me know!

gavanderhoorn:
@fmauch: did you uncheck `allow edits by maintainers`? I can't push some fixup commits.

fmauch:
> @fmauch: did you uncheck `allow edits by maintainers`? I can't push some fixup commits.

It seems to be checked.

---

Edit: There might be problems, as I forked my version from @dhled in the first place.

gavanderhoorn:
No, that should not matter.

fmauch:
image
This looks correct to me. Anything else I can try to follow up on this?

danielcranston:
Just wanted to say thanks @gavanderhoorn @fmauch for keeping up the momentum on getting `melodic-devel-staging` up to speed and making it more easily usable together with `ur_robot_driver`.

ipa-nhg:
@fmauch I tried to resolve the merge conflict and get the PR ready to merge but I can't push to your branch.

You can merge this https://github.com/fmauch/universal_robot/compare/moveit...ipa-nhg:fmauch-moveit?expand=1 or resolve the merge conflict by your self*.

----

(*) To solve the merge conflict

```
git clone git@github.com:fmauch/universal_robot -b moveit
cd universal_robot/
git pull --rebase git@github.com:ros-industrial/universal_robot melodic-devel-staging
```
The only merge conflict to resolve is the .travis.yml file that was [recently deleted](https://github.com/ros-industrial/universal_robot/pull/583).

This means that to solve it you have just to pick the "delete" option:

```
✗nhg-ThinkPad-T490(moveit|REBASE 1/14)/tmp/universal_robot>git mergetool 

This message is displayed because 'merge.tool' is not configured.
See 'git mergetool --tool-help' or 'git help config' for more details.
'git mergetool' will now attempt to use one of the following tools:
meld opendiff kdiff3 tkdiff xxdiff tortoisemerge gvimdiff diffuse diffmerge ecmerge p4merge araxis bc codecompare emerge vimdiff
Merging:
.travis.yml

Deleted merge conflict for '.travis.yml':
  {local}: deleted
  {remote}: modified file
Use (m)odified or (d)eleted file, or (a)bort? d
```

gavanderhoorn:
@ipa-nhg: I have https://github.com/fmauch/universal_robot/pull/20 open, so perhaps a rebase would be easier.

fmauch:
Merged https://github.com/fmauch/universal_robot/pull/20 and rebased to resolve conflicts.

gavanderhoorn:
@fmauch: somehow the commits from https://github.com/fmauch/universal_robot/pull/20 don't seem to have become part of this PR. Could you verify?

Did you perhaps `rebase` your local copy without merging the new changes and then force-pushed over the remote?

fmauch:
> @fmauch: somehow the commits from [fmauch#20](https://github.com/fmauch/universal_robot/pull/20) don't seem to have become part of this PR. Could you verify?
> 
> Did you perhaps `rebase` your local copy without merging the new changes and then force-pushed over the remote?

xxx indeed I think, that was the case, sorry!

gavanderhoorn:
You should be able to get the changes from [gavanderhoorn/universal_robot@moveit_fmauch_fixups](https://github.com/gavanderhoorn/universal_robot/tree/moveit_fmauch_fixups) and merge locally .

fmauch:
Now that should be better.

However, maybe before merging this, we recently had a related issue in the driver issue (https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/471) By default, the `move_group.launch` file [reloads the description](https://github.com/fmauch/universal_robot/blob/1b54d037e584dd23611e0b685bbabd860e30f806/ur10e_moveit_config/launch/move_group.launch#L40).

This would effectively overwrite a calibrated robot description when using these default packages with a calibrated robot description. I am unsure whether this is considered a problem or not.

gavanderhoorn:
All ROS nodes should be using the same description -- unless the user has specifically configured things to be different.

So yes, this is a problem.

However it should be easily solved: either we stop the MoveIt config from uploading the description. Or we change them to use the `load_X.launch` files, instead of what they do now.

The former would mean *something else* will be responsible for making sure the description is present. The latter would mean more custom code in the MoveIt cfgs.

gavanderhoorn:
<del>Hm. Seems the involved files are still trying to load limited variants and whatnot. This needs some work in any case ..</del>

Ignore: wrong branch.

fmauch:
> The former would mean _something else_ will be responsible for making sure the description is present. The latter would mean more custom code in the MoveIt cfgs.

The latter would also mean that the default moveit packages could not be used with a calibrated robot (although it is a good idea to create an own moveit config, anyway, questions will arise, I guess).

In which situation would you like the moveit config to load the description? Both, `ur_gazebo` and the `ur_robot_driver` (and I would expect all other existing drivers out there, as well) load the description together with the hardware abstraction.

Also, this is probably related to #585 

---

Edit: Also, when using the `load...launch` files inside the moveit packages (or using the custom macros on the `load...launch` files as suggested in #585) will only hide away the fact, that we will be loading the description twice. And as soon as somebody adapts the robot description and loads that with the driver and not changing this for the moveit config, there will be problems. So in my opinion disabling to load the description by default isn't a bad idea, anyway.

gavanderhoorn:
Unfortunately we'll probably need another rebase of this again.

fmauch:
I'll see when I can squeeze it in. Tonight, latest.

gavanderhoorn:
No problem. Thanks for the contribution already.

danielcranston:
So in my understanding this PR now does what it sets out to do.

There might still be the issue of "`move_group.launch` files [reloading the description](https://github.com/fmauch/universal_robot/blob/1b54d037e584dd23611e0b685bbabd860e30f806/ur10e_moveit_config/launch/move_group.launch#L40)" (as @fmauch pointed out above). IMO that's a topic for another issue/PR though, and this can be merged.

Is this in line with your thinking as well @gavanderhoorn ?

fmauch:
Before this gets forgotten again until the next WRID: @gavanderhoorn any more comments on this?

v4hn:
In response to [@fmauch's comment above](https://github.com/ros-industrial/universal_robot/pull/538#issuecomment-930073835): We just merged https://github.com/ros-planning/moveit/pull/3065 in MoveIt. Loading the description by default was an upstream error.

It seems this *did* get forgotten again over winter, so maybe it's time spring reinvigorates this stale pr? ;)
Judging by [this comment](https://github.com/ros-industrial/universal_robot/issues/573) it seems this is the last bit missing to stop advocating people to use @fmauch's branch?

ur10e_moveit_config/launch/moveit_rviz.launch
```diff
@@ -0,0 +1,15 @@
+<launch>
+
+  <arg name="debug" default="false" />
+  <arg unless="$(arg debug)" name="launch_prefix" value="" />
+  <arg     if="$(arg debug)" name="launch_prefix" value="gdb --ex run --args" />
+
+  <arg name="rviz_config" default="" />
```
RobertWilbrandt: This is currently missing the previous `config:=true` behavior of loading `$(find ur10e_moveit_config)/launch/moveit.rviz`. I personally would like defaulting this to that config, but this would invert the behavior when providing no argument at all (is there any reason why that was chosen? Just starting an empty rviz doesn't seem too useful).

fmauch:
re-opening to re-trigger CI

ur10e_moveit_config/launch/moveit_rviz.launch
fmauch: I agree, but let's keep the file's behavior to what it was on the kinetic-devel branch. I'll adapt to get the false/true behavior similar to the kinetic-devel branch.

Edit: As we are changing a lot in this PR and we'll have to update the README anyway, I've changed that in 51673f9 according to your suggestion.

ur10_moveit_config/package.xml
```diff
@@ -24,10 +24,12 @@
...
   <exec_depend>robot_state_publisher</exec_depend>
+  <exec_depend>tf2_ros</exec_depend>
   <exec_depend>xacro</exec_depend>
   <depend>ur_description</depend>
 
   <test_depend>roslaunch</test_depend>
   <buildtool_depend>catkin</buildtool_depend>
-  
+
+  <exec_depend>trac_ik_kinematics_plugin</exec_depend>
```
RobertWilbrandt: Why is this down here?

ur10_moveit_config/launch/planning_context.launch
```diff
@@ -1,20 +1,16 @@
...
 
   <!-- The name of the parameter under which the URDF is loaded -->
   <arg name="robot_description" default="robot_description"/>
 
   <!-- Load universal robot description format (URDF) -->
-  <group if="$(arg load_robot_description)">
-    <param unless="$(arg limited)" name="$(arg robot_description)" command="$(find xacro)/xacro --inorder '$(find ur_description)/urdf/ur10_robot.urdf.xacro'" />
-    <param if="$(arg limited)" name="$(arg robot_description)" command="$(find xacro)/xacro --inorder '$(find ur_description)/urdf/ur10_joint_limited_robot.urdf.xacro'" />
-  </group>
+  <param if="$(arg load_robot_description)" name="$(arg robot_description)" command="xacro  '$(find ur_description)/urdf/ur10.xacro'"/>
```
RobertWilbrandt, line 9: Why does this not just include the corresponding `load_*.launch`file?

gavanderhoorn: Because that would take this file further away from what is generated by the MSA, which would increase maintenance effort.

See https://github.com/ros-industrial/fanuc/pull/170 (and https://github.com/ros-industrial/fanuc/pull/208 for the replacement) for a PR which first tried to do what you suggest here (for a different OEM / in a different repository, but I had the same idea).

gavanderhoorn:
@fmauch: the last few commits you've pushed are all attributed to an `Ubuntu <ubuntu@rosmelodic.lxd>` user, instead of your own account.

ur10_moveit_config/package.xml
gavanderhoorn: Because it was an addition to the dependencies the MSA generates.

Keeping things separate like this simplifies reviews when the configuration package gets updated.

fmauch: But wouldn't it be nice to make sure that all dependencies are listed and that list is complete? If this is correctly updated before the PR (which IMHO makes sense to expect), reviews should still be nice and clean, right?

fmauch:
Should we address #431 also in this PR? @gavanderhoorn @RobertWilbrandt opinions?

RobertWilbrandt:
> Should we address #431 also in this PR? @gavanderhoorn @RobertWilbrandt opinions?

I would prefer to integrate this here, merging completely reworked moveit configs that don't follow this standard doesn't seem consistent to me.

fmauch:
I've tested this PR on real hardware. For using it with gazebo, @RobertWilbrandt is currently working on a fix of the gazebo xacro files.

As I am the author of this PR, it would be good if another maintainer could review this PR before merging.

ur10e_moveit_config/CHANGELOG.rst
```diff
@@ -0,0 +1,63 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package ur10_moveit_config
```
RobertWilbrandt:
```suggestion
Changelog for package ur10e_moveit_config
```

RobertWilbrandt (changes requested):
I think this is now at a stage that it can get merged:
- The demo.launch files work for each robot type
- Planning and executing trajectories using `ur_gazebo` works with each robot type (i got `PATH_TOLERANCE_VIOLATED` errors for ur16e when planning wtih full velocity, but i don't think this is caused by this PR)
- `catkin_lint -W2 universal_robot` only shows 3 warnings, and those are in `ur_kinematics` and not related to this PR
- `rosdep` has no problems with unknown dependencies
- I only found the one remaining copy-paste error while going through the commits (and checking for wrong robot types in each config repo)
- I tested the version on a ur3e, ur5e and a ur10. I could plan and execute trajectories without problems. 

I will aprove this after this last suggestion is fixed.

RobertWilbrandt (approved):

fmauch:
Merging, as the failed testing build on noetic seems to be a buildfarm glitch. It complains about a missing `ros-noetic-moveit-ros-visualization` package, which is indeed not part of the [package list](http://packages.ros.org/ros-testing/ubuntu/dists/focal/main/binary-amd64/Packages) atm. as `ros-noetic-moveit-ros`, `ros-noetic-moveit-ros-planning-interface` and `ros-noetic-moveit-ros-visualization` are missing, I suspect, this is due to [failing builds](https://build.ros.org/job/Nbin_uF64__moveit_ros_planning_interface__ubuntu_focal_amd64__binary/) of `ros-noetic-moveit-ros-planning-interface`.

PR #542, opened 2020-10-08
[melodic-devel-staging] Remove ur_msgs package for the Metapackage description

ipa-nhg:
as per subject..

### DIFF
--- a/universal_robots/package.xml
+++ b/universal_robots/package.xml
@@ -29,7 +29,6 @@
   <exec_depend>ur_description</exec_depend>
   <exec_depend>ur_gazebo</exec_depend>
   <exec_depend>ur_kinematics</exec_depend>
-  <exec_depend>ur_msgs</exec_depend>
 
   <export>
     <metapackage/>


universal_robots/package.xml
```diff
@@ -29,7 +29,6 @@
   <exec_depend>ur_description</exec_depend>
   <exec_depend>ur_gazebo</exec_depend>
   <exec_depend>ur_kinematics</exec_depend>
-  <exec_depend>ur_msgs</exec_depend>
```
gavanderhoorn, line 32: Would I be correct if I believe you do this because the package is now hosted in its own repository instead of in this one?

Metapackages can be used to install "logical sets of packages", not just those which happen to be hosted by the same repository (like `desktop` and `desktop_full` fi). While there is nothing in `ros-industrial/universal_robot` which uses `ur_msgs` directly any more, we could consider still keeping it as a dependency here as it could make sense for users who install the `universal_robots` package to also get `ur_msgs` installed.

We could also decide that it should be the responsibility of packages actually depending on `ur_msgs` to make sure it gets installed.

And both options also do not need to be mutually exclusive.

ipa-nhg: To me make not much sense to keep the dependency here. As you pointed, there is nothing in the repo that needs the `ur_msgs` package. The package which needs the msgs should list it as dependency, as the [ur_robot_driver](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/package.xml) does.

I also understand your argument about the "set of packages", where the universal_robot set can represent the common basis for the UR drivers. But to be honest, the problem came to me when I wanted to test the simulation, I cloned the repo to my workspace and called `rosdep install from-src ...` . The command fails because of the ur_msgs dependency then, from a practical point of view and to avoid this kind of issues for future distros/releases, I would like to remove the dependency.

gavanderhoorn: > I cloned the repo to my workspace and called `rosdep install from-src ...` . The command fails because of the ur_msgs dependency then,

Which OS? `ur_msgs` has been released into Melodic, so I would expect `rosdep` to be able to resolve it.

ipa-nhg: actually Melodic, probably a `rosdep update` fix the problem. 

But still this can happen for every new distro, if the `universal_robot` repo is not released then, very likely, the `ur_msgs` neither. Then the users will need to clone the source of both repositories to the workspace, information that should be part of the readme.

gavanderhoorn: > Then the users will need to clone the source of both repositories to the workspace, information that should be part of the readme.

Of course.

And we'll actually provide a `.rosinstall` file.

gavanderhoorn (approved):
Let's go for the simpler setup and remove it here.

gavanderhoorn:
Thanks @ipa-nhg :+1:

PR #583, opened 2021-09-29
Switch to Github Actions for CI

gavanderhoorn:
As per subject.

Still using `industrial_ci`, but now hosted by Github Actions.

Fairly minimal `matrix` build of Melodic and Noetic on Bionic and Focal. Might add Xenial+Kinetic back later (for Ubuntu ESM).

Runs here: [gavanderhoorn/universal_robot/actions](https://github.com/gavanderhoorn/universal_robot/actions).

PR #584, opened 2021-09-29
readme: use GHA badges

gavanderhoorn:
As per subject.

I forgot to include this in #583.

Note: these links won't work until the workflows are actually run here in this repository. But that won't happen until we make the branch containing them the default one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # Universal Robot
 
 [![Build Status](http://build.ros.org/job/Kdev__universal_robot__ubuntu_xenial_amd64/badge/icon)](http://build.ros.org/job/Kdev__universal_robot__ubuntu_xenial_amd64)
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Closed/re-opened to trigger actions.

gavanderhoorn:
I specifically didn't do that (ie: point to `melodic-devel-staging`), as the intent is for `melodic-devel-staging` to be merged into `melodic-devel`.

Tbh, the badges not showing anything for `melodic-devel-staging` is not really a problem. We know CI works, the badges are just a cosmetic problem which will solve itself in the near future.

---

Edit: thanks for the attention to detail though.

ipa-nhg (approved):
LGTM! 👍🏽 

Many thanks!

PR #587, opened 2021-09-29
readme: there are no Kinetic dev jobs any more

gavanderhoorn:
As per subject.


ipa-nhg (approved):
LGTM! 👍🏽 

I was about to propose the same change now... Thanks a lot!

PR #588, opened 2021-09-29
ur_description: Use the per-model xacro files for correct robot name

### LINKED ISSUE #585: description: Should the load_urXX.launch files use the robot-specific macros?
As risen in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/471 the robot model loaded by the `load_urXX.launch` files of `ur_description` create a robot called `ur`:

https://github.com/ros-industrial/universal_robot/blob/0101cae607c9a8e7fc72934849eb4f2343e58ded/ur_description/urdf/ur.xacro#L2

When used with the moveit_config packages (see #538), this name clashes with the SRDF files configured there, when the moveit startup should not re-load the description. It might be a good idea to use the named macros inside the `load_urXX.launch` files.

Is there anything speaking against this?

fmauch:
With the common xacro file the robot name will always be "ur".
However, the moveit config packages expect the robot name to be e.g. "ur10e".
This commit will use the macros containing the model name already.

This fixes #585 and should help in #538.

### DIFF
--- a/ur_description/launch/load_ur.launch
+++ b/ur_description/launch/load_ur.launch
@@ -11,6 +11,8 @@
   <arg name="safety_pos_margin" default="0.15" doc="The lower/upper limits in the safety controller" />
   <arg name="safety_k_position" default="20" doc="Used to set k position in the safety controller" />
 
+  <arg name="robot_model" />
+
   <!-- Load the top-level (ie: stand-alone and complete) xacro for the UR
        variant defined by the set of yaml parameter files (so to load a UR5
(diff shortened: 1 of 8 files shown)


gavanderhoorn (approved):
I'm not too thrilled about the `robot_model` argument (should it be called 'variant'?), as this adds yet another argument to an already long list.

But it's not exposed to users (they probably don't want to use the base `load_ur.launch`) and it does solve the problem.

gavanderhoorn:
Thanks @fmauch :+1:

PR #607, opened 2022-08-18
Fixed ur5 base height

BobbyCephy:
The visual and collision mesh of the base of the ur5 has to be moved 3 mm upwards in order to close the gap between the shoulder and align its bottom in the xy-plane. Otherwise the ur5 collides with the surface it is directly placed on.

**Before**
before

**After**
after


fmauch (approved):
Thanks for spotting this! This file has been like that for a long time! Maybe that was caught in the URDF in the past, but since we attach the mesh to the link's origin at least since we merged all descriptions, this is obviously wrong.

PR #609, opened 2022-11-01
All parameters can be set in top-level xacros

BobbyCephy:
All parameters can be passed by arguments to top-level xacros. Otherwise default macro values are used.

### DIFF
--- a/ur_description/urdf/ur.xacro
+++ b/ur_description/urdf/ur.xacro
@@ -3,19 +3,20 @@
    <!-- Import main macro -->
    <xacro:include filename="$(find ur_description)/urdf/inc/ur_macro.xacro"/>
 
-   <!-- Parameters -->
-   <xacro:arg name="robot_model" default="ur5"/>
-   <xacro:arg name="prefix" default=""/>
-   <xacro:arg name="joint_limit_params" default="$(find ur_description)/config/$(arg model)/joint_limits.yaml"/>
-   <xacro:arg name="kinematics_params" default="$(find ur_description)/config/$(arg model)/default_kinematics.yaml"/>
-   <xacro:arg name="physical_params" default="$(find ur_description)/config/$(arg model)/physical_parameters.yaml"/>
-   <xacro:arg name="visual_params" default="$(find ur_description)/config/$(arg model)/visual_parameters.yaml"/>
-   <!-- Legal values:
-         - hardware_interface/PositionJointInterface
-         - hardware_interface/VelocityJointInterface
-         - hardware_interface/EffortJointInterface
-   -->
+   <!-- Type/series of used UR robot (ur3, ur3e, ur5, ur5e, ur10, ur10e, ur16e) -->
+   <xacro:arg name="robot_model" default="$(arg robot_model)"/>
+   
+   <!-- Transmission hardware interfaces:               hardware_interface/PositionJointInterface
+                                                        hardware_interface/VelocityJointInterface
+                                                        hardware_interface/EffortJointInterface  -->
    <xacro:arg name="transmission_hw_interface" default="hardware_interface/PositionJointInterface"/>
+   
+   <!-- Parameters -->
+   <xacro:arg name="prefix" default=""/>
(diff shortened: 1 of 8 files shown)


ur_description/urdf/ur10.xacro
```diff
@@ -17,6 +17,6 @@
 
     Refer to 'inc/ur_macro.xacro' for more information.
   -->
-  <xacro:include filename="$(find ur_description)/urdf/inc/ur10_macro.xacro"/>
-  <xacro:ur10_robot prefix="" />
+  <xacro:arg name="model" default="10"/>
+  <xacro:include filename="$(find ur_description)/urdf/ur.xacro"/>
```
fmauch: We currently have a theme in place that we don't want to include an xacro file for code contribution, but only as an include that provides macros. Thus, the macro is invoked instead.

In fact, if I understand that correctly, this file doesn't do anything, since it's not calling any macro.

Furthermore this file is intended as an example only, as stated in the comment above.

ur_description/urdf/ur.xacro
```diff
@@ -1,15 +1,16 @@
...
    <xacro:include filename="$(find ur_description)/urdf/inc/ur_macro.xacro"/>
 
-   <!-- parameters -->
-   <xacro:arg name="joint_limit_params" default=""/>
-   <xacro:arg name="kinematics_params" default=""/>
-   <xacro:arg name="physical_params" default=""/>
-   <xacro:arg name="visual_params" default=""/>
-   <!-- legal values:
+   <!-- Parameters -->
+   <xacro:arg name="model" default="5"/>
```
fmauch:
```suggestion
   <xacro:arg name="robot_model" default="ur5"/>
```
As we nowadays do add the `robot_model` argument to the robot's name, we should call the argument the same here and add the `ur` prefix to its possible values. Also, possible values should probably be added as a comment above.

ur_description/urdf/ur.xacro
```diff
@@ -1,15 +1,16 @@
...
+   <xacro:arg name="prefix" default=""/>
+   <xacro:arg name="joint_limit_params" default="$(find ur_description)/config/ur$(arg model)/joint_limits.yaml"/>
+   <xacro:arg name="kinematics_params" default="$(find ur_description)/config/ur$(arg model)/default_kinematics.yaml"/>
+   <xacro:arg name="physical_params" default="$(find ur_description)/config/ur$(arg model)/physical_parameters.yaml"/>
+   <xacro:arg name="visual_params" default="$(find ur_description)/config/ur$(arg model)/visual_parameters.yaml"/>
```
fmauch:
```suggestion
   <xacro:arg name="joint_limit_params" default="$(find ur_description)/config/$(arg model)/joint_limits.yaml"/>
   <xacro:arg name="kinematics_params" default="$(find ur_description)/config/$(arg model)/default_kinematics.yaml"/>
   <xacro:arg name="physical_params" default="$(find ur_description)/config/$(arg model)/physical_parameters.yaml"/>
   <xacro:arg name="visual_params" default="$(find ur_description)/config/$(arg model)/visual_parameters.yaml"/>
```

fmauch (commented):
I generally like the proposed idea, but I'd suggest not touching the `urX.xacro` standalone files with this. They are meat as standalone example files only.

I'll have to think again, whether these changes would collide with the issues discussed in #592, but at a first glance, I think this should be alright.

Note: Given on the timing, this PR might have to be re-added once `melodic-devel-staging` got merged, see #617.

ur_description/urdf/ur10.xacro
BobbyCephy: A specific urX.xacro includes the generic ur.xacro file with the default robot_model arg changed only. All default args in ur.xacro can be changed in urX.xacro as well, since the are forwarded. Only the generic macro in ur_macro.xacro is included and called by ur.xacro.
A urX_macro.xacro might call the generic macro with robot_model only as well in an adjusted ur_macro.xacro.

PR #615, opened 2022-11-08
Make ur_gazebo robot name consistent with ur_description

RobertWilbrandt:
This fixes the different urdf robot names between the real hardware and ur_gazebo and thereby eliminates a moveit startup error that complains about a name difference between the urdf and srdf.


fmauch (approved):
LGTM, thanks!

Issue #613, opened 2022-11-07
Future of ur_kinematics plugin


fmauch:
Currently, the main thing holding us back from doing a binary release is the ur_kinematics plugin. The current identified problems are:

* It uses a deprecated API (See #476), while #476 makes the move_group crash on startup
* URKinematics currently only supports UR3, UR5, UR10, while all the four e-Series robots are missing
* The URKinematics plugin uses the uncalibrated robot model, so it won't produce accurate results

I am not sure what the scope / impact of this would be. With many alternative kinematics plugins available, I'm asking myself whether it would be worthwile addressing this. For the sake of no longer holding things back I would currently vote for merging melodic-devel-staging into melodic-devel without fixing that package and ignoring it in the first melodic/noetic release.

If you think that it would be beneficial having this available or you even would like to contribute to fixing this, please let us know.

fmauch:
Update: We've updated the package so that it supports all 7 robot models, but the other two points remain valid in my opinion.

simonschmeisser:
Just as a quick update, I opened a PR to fix the crash in #476  (I could also open a new MR directly)

For us the analytical IK is valuable as it tends to be about a factor 1000 faster than the numerical ones. Since we can easily have 50 000 IK requests in our bin picking setups this adds up

PR #616, opened 2022-11-08
Update ur kinematics

### LINKED ISSUE #460: Add support for UR16e
Recently released.

`ur_support` should be extended to include meshes, config, launch files and xacro macro.

fmauch:
In #613 I collected a couple of reasons why I think that ur_kinematics is currently flawed. This PR addresses a couple of them.

* Adds parameters for all robot models
* Removes the `kinematics_solver_attempts` as suggested in #474
* Corrects plugin name in `kinematics.yaml` files

Since I still see issues with using this plugin due to its problems regarding robot calibration I added a README to inform users about this concern.

This closes #460

### DIFF
--- a/ur_kinematics/src/ur_kin.cpp
+++ b/ur_kinematics/src/ur_kin.cpp
@@ -13,6 +13,16 @@ namespace ur_kinematics {
     }
     const double PI = M_PI;
 
+    //#define UR16e_PARAMS
+    #ifdef UR16e_PARAMS
+    const double d1 =  0.1807;
+    const double a2 = -0.4784;
+    const double a3 = -0.36;
+    const double d4 =  0.17415;
+    const double d5 =  0.11985;
+    const double d6 =  0.11655;
+    #endif
+
     //#define UR10_PARAMS
     #ifdef UR10_PARAMS
     const double d1 =  0.1273;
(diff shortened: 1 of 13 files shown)


fmauch:
BTW: The DH parameters are extracted from https://www.universal-robots.com/articles/ur/application-installation/dh-parameters-for-calculations-of-kinematics-and-dynamics/

ur_kinematics/src/ur_kin.cpp
```diff
@@ -23,6 +33,16 @@ namespace ur_kinematics {
     const double d6 =  0.0922;
     #endif
 
+    //#define UR10e_PARAMS
+    #ifdef UR10e_PARAMS
+    const double d1 =  0.1807;
+    const double a2 = -0.627;
```
RobertWilbrandt:
```Suggestion
    const double a2 = -0.6127;
```
Copy-paste error

RobertWilbrandt (changes requested):

RobertWilbrandt (approved):

gavanderhoorn:
Was this merged with a squash-merge?

Please only use those for single-commit PRs.

fmauch:
I would have squashed the PR anyway, so we simply skipped that step.

gavanderhoorn:
Why? All 4 commits make sense to me to be separate commits.

Thematically they don't belong together, so better to keep them as separate commits.

fmauch:
OK, I'll keep that in mind in future.

PR #617, opened 2022-11-08
Melodic devel staging

fmauch:
I believe that melodic-devel-staging is elaborate enough to be merged into melodic-devel.


melodic.ignored
```diff
@@ -0,0 +1 @@
+ur_kinematics
```
gavanderhoorn: Wouldn't this have to be added to the release repository?

fmauch: I was going after https://answers.ros.org/question/66545/to-ignore-some-pkgs-upon-release-by-bloom/?answer=66548#post-id-66548 but you might be right. I've never done that.

RobertWilbrandt (approved):
With #619 merged all breaking changes should now be done and there is no more reason to delay a release.

fmauch:
Not deleting the source branch, since there are many references on that.

PR #628, opened 2023-01-03
Added missing 's' in installation command

jc-cr:
's' character was missing from the installation command.


jc-cr:
Seems like the Kinect and other Melodic branches are also missing the 's' character.

fmauch (approved):
Thanks for spotting this! The metapackage was renamed quite a while ago and this got overseen when updating the README.

PR #629, opened 2023-01-26
ci: bump checkout and cache

gavanderhoorn:
As per subject.

The currently used versions are deprecated and produce a number of warnings (and soon errors).


gavanderhoorn:
I'd recommend squash-merging this, as it's essentially a single change.

fmauch (approved):
Thanks for updating this

PR #635, opened 2023-04-26
add "S" to "Boost_INCLUDE_DIR" line 26

pkoprov:
This is addressing an issue where boost is looking for DIRS

### DIFF
--- a/ur_kinematics/CMakeLists.txt
+++ b/ur_kinematics/CMakeLists.txt
@@ -23,7 +23,7 @@ catkin_package(
 ## Build ##
 ###########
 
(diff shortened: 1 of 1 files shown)


pkoprov:
it actually partially fixes issue https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/432

RobertWilbrandt (approved):
Thanks for the PR (and sorry for the long delay), this looks like an oversight and the correct way to include boost headers: [Documentation](https://cmake.org/cmake/help/latest/module/FindBoost.html).

PR #659, opened 2023-10-10
Added prefix argument to ur.xacro

Doomerdinger:
There are some use cases where being able to specify the prefix in the non-macro xacro file is useful.

Adding this should have no impact on any existing use cases.

I can add this argument to the other "concrete" xacro files, but I believe in nearly all cases where this is relevant the user will favor using the generic file anyway.

As a side note, this argument is provided in the non-macro file of the [ROS2 description repo](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/75b5c69b79de206bbf646b681ce345059e01c12b/urdf/ur.urdf.xacro#L13C11-L13C11).

### DIFF
--- a/ur_description/urdf/ur.xacro
+++ b/ur_description/urdf/ur.xacro
@@ -18,10 +18,11 @@
    <xacro:arg name="safety_limits" default="false"/>
    <xacro:arg name="safety_pos_margin" default="0.15"/>
    <xacro:arg name="safety_k_position" default="20"/>
+   <xacro:arg name="prefix" default=""/>
 
    <!-- arm -->
    <xacro:ur_robot
-     prefix=""
+     prefix="$(arg prefix)"
      joint_limits_parameters_file="$(arg joint_limit_params)"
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
> There are some use cases where being able to specify the prefix in the non-macro xacro file is useful.

Could you give an example where this would be useful?

As you already write yourself, this `.xacro` file is supposed to be a top-level one and is not meant for composition. I'm having a hard time coming up with a use-case for a prefix.

Doomerdinger:
An example would be when using external tooling to compose URDFs, but still wanting to utilize existing xacro files alongside/within those tools.
By virtue of what those tools are attempting, they essentially must be able to parse URDF properly, but may not support parsing/resolving xacro macros.
In such cases, it can be useful to have a top-level xacro file that essentially wraps the macro more or less 1:1, so that `xacro` itself can be called on that top-level file and the resulting valid URDF file fed into the other tool.

PR #677, opened 2024-04-08
Fix default calibration file for UR30

fmauch:
Apparently that went wrong when adding this, as we only changed the DH parameters without updating the hash.


RobertWilbrandt (approved):
This looks good to me, thank!
```
[...]
[ INFO] [1712579871.220693856]: Checking if calibration data matches connected robot.
[ INFO] [1712579872.223066645]: Calibration checked successfully
[...]
```

PR #685, opened 2024-09-06
[ur_gazebo] Add possibility to specify transmission_interface and position-controllers

fmauch:
As suggested in https://github.com/moveit/moveit_tutorials/pull/753#issuecomment-1410513014


RobertWilbrandt (approved):
This looks good, it might be nice to have some documentation about when to use which but it's good to be able to use this now.

Issue #195, opened 2015-04-10
[C API] Scons compile error - set connected() is protected 


2Reeko:
Hi,

I've been doing the installation step 1 of ur_c_api_bringup and I did everything upto the compile step. Just putting the specified files into a directory on the control box and creating the symolic links for the compiler. I had an issue earlier with the scons compiler not being able to read the simple_message .h or .cpp files but it was just the character encoding that it didn't recognise. Now when I try to compile the control server it stops early with this error.

```
include/simple_message/socket/simple_socket.h: In member function 'void ur::URControllerInterface::sendAndReceiveMessages()':
include/simple_message/socket/simple_socket.h:253: error: 'virtual void industrial::simple_socket::SimpleSocket::setConnected(bool)' is protected
src/ur_controller_iface.cpp:51: error: within this context
scons: *** [build/test/src/ur_controller_iface.o] Error 1
scons: building terminated because of errors.
```

Seems that both connected_ and setconnected() are protected in the simple_socket.h. This throws an error when the compiler hits this section of ur_controller_iface specifically the sendandreceivemessages function;

```
  if(ur_state.sequence - latest_cmd_seq >= CMD_TIMEOUT) {
    // printf("TIMEOUT\n");
    connection->setConnected(false);
```

Any ideas on how to get around this? or have I just missed something I should have done?

Thanks,

Christian

gavanderhoorn:
I remember something about some methods being protected in `simple_message` and @kphawkins running into this as well some time ago. @shaun-edwards: do you remember how/if this was resolved?

@2Reeko: could you specify which particular version of the C-API driver you are attempting to use? The branch on this repository is rather old, and has not been kept up-to-date with the changes in the [gt-ros-pkg/universal_robot](https://github.com/gt-ros-pkg/universal_robot) fork.

2Reeko:
@gavanderhoorn I followed the link from http://wiki.ros.org/ur_c_api_bringup which appears to be the same as the link you posted. Not sure how to find the version 'number'? 

branch: hydro-c-api-ur10-moveit
latest commit: 8fe5153 cfa
authored: kphawkins dec 5, 2014

shaun-edwards:
[...]

PR #689, opened 2024-10-30
Fix masses of robot links

VinDp:
Changed the masses to match the actual specifications, as done for ROS2 in [#187](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/187).

### DIFF
--- a/ur_description/config/ur10/physical_parameters.yaml
+++ b/ur_description/config/ur10/physical_parameters.yaml
@@ -14,13 +14,13 @@ offsets:
 
 inertia_parameters:
   base_mass: 4.0  # This mass might be incorrect
-  shoulder_mass: 7.778
-  upper_arm_mass: 12.93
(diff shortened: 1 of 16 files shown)


VinDp:
I also added the changes from [#195](https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/pull/195) to be on pair with the ROS 2 version. 
While doing so I also noticed that there were some parts that could maybe be cleaned up, like in [physical_parameters.yaml](https://github.com/ros-industrial/universal_robot/blob/cce1f0e55b265b8440a333a4c39caae56a6e0b16/ur_description/config/ur30/physical_parameters.yaml#L26) of the `ur30`. Should I include that in this PR or is it better to leave it for a different one?

urfeex (approved):
Finally got around to testing this. Thank you @VinDp

PR #692, opened 2025-03-27
Add support for ur7e and ur12e

urfeex:
Add description, moveit and gazebo packages for UR7e and UR12e. As they are basically a UR5e and UR10e with updated internals, I copied and symlinked basically everything.

Once could probably argue, whether symlinking more instead of copying would make things easier to use / maintain.


gavanderhoorn:
Do symlinks work on Windows?

urfeex:
> Do symlinks work on Windows?

Do you think that would be relevant in the context of ROS 1?

ur_gazebo/launch/inc/load_ur7e.launch.xml
```diff
@@ -0,0 +1,19 @@
...
+
+  <!--Common parameters -->
+  <arg name="transmission_hw_interface" default="hardware_interface/EffortJointInterface" doc="The hardware_interface to expose for each joint in the simulated robot (one of: [PositionJointInterface, VelocityJointInterface, EffortJointInterface])"/>
+  <arg name="safety_limits" default="false" doc="If True, enable the safety limits controller"/>
+  <arg name="safety_pos_margin" default="0.15" doc="The lower/upper limits in the safety controller" />
+  <arg name="safety_k_position" default="20" doc="Used to set k position in the safety controller" />
+
+  <!-- Use common launch file and pass all arguments to it -->
+  <include file="$(dirname)/load_ur.launch.xml" pass_all_args="true">
+    <arg name="robot_model" value="ur5e" />
```
urmahp:
```suggestion
    <arg name="robot_model" value="ur7e" />
```

urmahp (changes requested):
Overall looks good. I only found one small adjustment.

urfeex:
I went back to copying files instead of symlinks, also because the robot's names would be wrong in the robot descriptions, otherwise.

urmahp (approved):

PR #700, opened 2025-05-06
ci: pin runner image to Noble

gavanderhoorn:
As per title.

Keeps us in control of which runner image is used to run actions on.

We don't typically like *moving targets* that can be changed without notification.


urfeex (approved):
Fine with me!

